PR #18, opened 2015-01-12
portal bug due to orphaned notificaitons

ying-pbrc:
null reference exception occurs on portal page when a service request with notifications was  deleted.


amcates:
Ha, funny we both found this today.  Thanks!

ying-pbrc:
I actually found this last week, but didn’t want to bother you with too many issues J

PR #19, opened 2015-01-12
fix one time fees services issues

ying-pbrc:
1. one time fees services missing from the sub service request dialog in portal.
2. make sure one time fees services are not displayed twice on review page.


amcates:
The team is internally reviewing this and will provide feedback.

jleonardw9:
Thanks for pointing out this bug. I had to make a modification in the _otf_calender.html.haml file to account for the duplicates on the review page, but good catch in the portal view. I'm going to delete this pull request and then give you credit in the one I'm going to submit.

ying-pbrc:
No problem. Thanks.

amcates:
Fixed in https://github.com/sparc-request/sparc-request/pull/25

PR #20, opened 2015-01-12
fix otf service report - missing program services

ying-pbrc:
missing program level services if Core is not being used
d

### DIFF
--- a/lib/tasks/one_time_fee_report.rake
+++ b/lib/tasks/one_time_fee_report.rake
@@ -66,12 +66,12 @@ namespace :data do
           total = 0
           if li.service.is_one_time_fee? && (li.created_at.to_date > 2012-03-01)
             if li.fulfillments.empty?
-              row = [protocol_id, service_request_id, status, short_title, pi, provider.abbreviation, program.abbreviation, core == "" ? "" : core.abbreviation, owner, li.service.name, (li.in_process_date.to_date rescue nil), (li.complete_date.to_date rescue nil), "null", "null", "null", "null"]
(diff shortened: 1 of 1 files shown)


amcates:
2 issues with this request.

1. Wouldn't core.blank? or core.empty? be a better option than core == ""
2. The loop at the bottom will duplicate data.  You need to loop over programs with no core and programs with cores separately.

ying-pbrc:
Agree. Thanks! I will take care of it.

ying-pbrc:
Thanks for your feedback. I’ve changed core == “” to core.blank? as suggested.

 

After examining the bottom outer loop 

            provider.programs.each do |program|

and testing with various scenarios, I don’t think there is a possibility of getting a duplicate since the two inner loops:  

  program.sub_service_requests.each do |ssr|

    build_one_time_fee_report(csv, ssr, provider, program, "")

 end

and  

  program.cores.each do |core|

    core.sub_service_requests.each do |ssr|

      build_one_time_fee_report(csv, ssr, provider, program, core)

    end 

 end

are mutually exclusive.  

            

However, I came across another issue when the split/notify is checked at the program level for a provider and each program has its own cores: Although there was no major data error in the report, the Core column are all blank. In my test case, the organization_id of the program/cores subservice requests (ssr) are all set to the program instead of their corresponding core. In order to fix this issue, I probably need to modify the build_one_time_fee_report method. Do you have any suggestions? I can make changes to our SPARC instance and delete this pull request if you prefer. 

 

Thanks,

Ying

amcates:
Sorry, I misread the pull request.  I thought the second loop was inside the first.  It should work as you mentioned.

I will have the team re-exam the pull request and accept it.

Thanks!  


We will also look in to what you mentioned about the split/notify issue.

ying-pbrc:
Thanks.

PR #25, opened 2015-01-14
Modification to fix by ying-pbrc

jleonardw9:
@ying-pbrc thanks for the heads up on this!


amcates:
Thanks @ying-pbrc .  Jason's fix is based off what you mentioned.

PR #36, opened 2015-02-09
Restricting editing of requests by status

jleonardw9:


app/models/sub_service_request.rb
```diff
@@ -264,14 +264,41 @@ def ctrc?
     self.organization.tag_list.include? "ctrc"
   end
 
+  # Can't edit a Nexus ssr if it's placed in an uneditable status
   def can_be_edited?
-    ['first_draft', 'draft', 'submitted', nil, 'obtain_research_pricing'].include?(self.status) ? true : false
```
amcates: So, does this mean that all other requests can be edited regardless of status.

jug2: Yes – no matter the status, all other requests can be edited at any point.

PR #60, opened 2015-03-13
added ability to run .where() queries on an object

robertjamesmiller:
Andrew and Greg,

Please review this pull request and let me know what you think of the implementation and the direction I'm heading with it. I've got more tests I'm planning to write. For example, would it be okay if I added ":title" to ProtocolShallow?

Thanks,
Rob


app/api/v1/helpers_v1.rb
```diff
@@ -12,11 +12,19 @@ def find_object(klass, id)
...
 
-  def find_objects(klass, ids)
+  def find_objects(klass, params)
     klass = klass.classify
 
-    if ids.any?
-      @objects = klass.constantize.where(id: ids)
+    if params[:ids].any?
+      @objects = klass.constantize.where(id: params[:ids])
+    elsif params[:query] && params[:query].length > 0 && params[:limit] == 1
```
goleksiak: I feel like setting a default of: 1 on the "optional :limit, type: Integer" might be preferable here. This appears to limit a custom query to a SQL LIMIT of: 1

spec/api/v1/identities/get_identity_by_ldap_uid_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+         params = { :depth => 'shallow',
+                    :limit => 1,
+                    :query => { ldap_uid: "" } 
+                  }
+          cwf_sends_api_get_request_for_resources_by_params('identities', params) 
+       end
+       it 'should respond with an error' do
+         parsed_body         = JSON.parse(response.body)
+         expect(parsed_body['identity']).to eq(nil)
+         expect(parsed_body['error']).to eq("Identity not found for query #<Hashie::Mash ldap_uid=\"\">")
```
goleksiak: Is there a reason that an object is being returned in the error message? Maybe just return "Identity not found for query"

robertjamesmiller: To provide query details to the client to help with debugging. For example, it will help when third party app end users report issues but are unable to provide adequate details; because this information will be in our Java log files. How does that sound?

app/api/v1/helpers_v1.rb
robertjamesmiller: I'll set the limit default to 1. The "limit" param is a placeholder for now since I only coded the ability to return a single object but it will be used once I/we have enhanced this query feature to support the returning of a collection of objects.

robertjamesmiller:
Greg,

I've updated this pull request; please review when convenient. We're now going to work on our Java client, which should give us good feedback.

Thanks,
Rob

robertjamesmiller:
Greg,

I updated this pull request to add HumanSubjectsInfo to the API.

Rob

robertjamesmiller:
Greg and Andrew,

Please accept and merge this pull request before making additional changes to the GO-API branch so as to avoid more merging conflicts.

Thank you,
Rob

.gitignore
```diff
@@ -7,10 +7,13 @@
 sr.sql
 sr_production.sql
 
+<<<<<<< HEAD
```
goleksiak: Looks like some merge conflicts still exist

app/api/v1/entities/full.rb
```diff
@@ -27,6 +27,20 @@ class IdentityFull < IdentityShallow
             :last_name,
             :ldap_uid
   end
+  
+  class HumanSubjectsInfoFull < HumanSubjectsInfoShallow
+    root 'human_subjects_info', 'human_subjects_info'
```
goleksiak: I believe the first root param should be pluralized. https://github.com/intridea/grape-entity#collection-exposure

app/api/v1/entities/full.rb
```diff
@@ -27,6 +27,20 @@ class IdentityFull < IdentityShallow
...
+
+    expose  :protocol_id,
+            :nct_number,
+            :hr_number,
+            :pro_number,
+            :irb_of_record,
+            :submission_type,
+            :irb_approval_date,
```
goleksiak: Dates should be iso8601 formatted. See https://github.com/sparc-request/sparc-request/blob/go-api/app/api/v1/entities/full.rb#L46

app/api/v1/entities/shallow.rb
```diff
@@ -35,6 +39,7 @@ class ProjectRoleShallow < Shallow
 
   class ProtocolShallow < Shallow
     root 'protocols', 'protocol'
+    expose :short_title
```
goleksiak: Shallow class object attributes should be limited to [:id, :callback_url]. Slippery slope

app/api/v1/helpers_v1.rb
```diff
@@ -20,16 +20,26 @@ def presenter(klass, depth)
...
-  def find_objects(klass, ids)
+  def find_objects(klass, params)
     klass = klass.classify
 
-    if ids.any?
-      @objects = klass.constantize.where(id: ids)
-    else
+    if params[:ids].any?
+      @objects = klass.constantize.where(id: params[:ids])
+    elsif params[:query] && params[:query].length > 0
```
goleksiak: Could be refactored to: params[:query].present?

app/api/v1/helpers_v1.rb
```diff
@@ -20,16 +20,26 @@ def presenter(klass, depth)
...
+      # identify invalid parameters (not found in the object)
+      invalid_query_parameters = params[:query].select {|key, value| !klass.constantize.column_names.include? key }
+      if invalid_query_parameters && invalid_query_parameters.length > 0
```
goleksiak: Could be refactored to: invalid_query_parameters.present?

app/api/v1/helpers_v1.rb
```diff
@@ -20,16 +20,26 @@ def presenter(klass, depth)
...
+        error!("#{klass} query #{params[:query]} has the following invalid parameters: #{invalid_query_parameters.keys}", 400)
+      elsif params[:limit] == 1 # return only one object, the first that meets the query criteria
+        error!("#{klass} not found for query #{params[:query]}", 404) unless @object = klass.constantize.where(params[:query]).first
+      else # return all objects that meet the query criteria
+        @objects = klass.constantize.where(params[:query]).limit(params[:limit]) # a nil limit is ignored by ActiveRecord
```
goleksiak: We need a guard against SQL injection in params[:query]. ...hmmm

.gitignore
robertjamesmiller: fixed

app/api/v1/entities/full.rb
robertjamesmiller: fixed.

app/api/v1/entities/full.rb
robertjamesmiller: fixed.

app/api/v1/entities/shallow.rb
robertjamesmiller: It's a tradeoff because if we don't expose :short_title in Shallow then we're making a separate web service call for Full protocols and will receive 29 other columns we don't need, multiplied by the number of requested protocols. With that said, we can update our web service client to make a separate call for full protocols and wait until performance becomes an issue. How should we proceed: as currently coded or remove short_title?

Also, I think IdentityShallow should expose :email, :first_name, :last_name, and :ldap_uid because it seems like every scenario in which I'll be looking up an Identity, I'll be doing it so I can display and/or use all of those fields. What do you think?

app/api/v1/helpers_v1.rb
robertjamesmiller: refactored.

app/api/v1/helpers_v1.rb
robertjamesmiller: refactored.

amcates:
@goleksiak @robertjamesmiller  Thanks for working on this together.  Please let me know when it's ready for final review.

app/api/v1/helpers_v1.rb
robertjamesmiller: we're protected from SQL injection because Query must be of type Hash:
  params :custom_query do
    optional :limit, type: Integer
#    optional :order, type: String   
    optional :query, type: Hash
  end
end

And, there are tests in place (e.g., context 'query as type String is not valid, must be a Hash' do)

robertjamesmiller:
@amcates This pull request is ready for another review, thanks, and have a great weekend.

robertjamesmiller:
@amcates, @jayhardee9, and @goleksiak; pull request #119  applies significant changes to the API files and that will make merging this pull request more difficult.  after pull request #119 has been merged with master, if I delete this pull request and reapply my changes will you merge my changes before making any more API changes?

PR #69, opened 2015-04-09
suppress hidden field for is_one_time_fee when its value is true

robertjamesmiller:
Since last Fall in all of our environments, a service with Pricing Maps that have is_one_time_fee = true has them set to false on any "Save" call within Catalog manager for that service. The issue appears to be that a hidden field with a value of 0 is always present whereas the disabled checkbox's "true" value is not present:
http://stackoverflow.com/questions/7730695/whats-the-difference-between-disabled-disabled-and-readonly-readonly-for-ht

Since, readonly checkboxes won't work for us in this scenario: http://stackoverflow.com/questions/155291/can-html-checkboxes-be-set-to-readonly?page=1&tab=votes#tab-top , I've removed the hidden field when is_one_time_fee = true.

Please let me know if this fix (for us) causes issues for your institution.

### DIFF
--- a/app/views/catalog_manager/services/_pricing_map.html.haml
+++ b/app/views/catalog_manager/services/_pricing_map.html.haml
@@ -95,7 +95,7 @@
         %tr.otf
           %th= t(:pricing_map_form)[:is_one_time_fee]
           %td.is_one_time_fee{colspan: 2}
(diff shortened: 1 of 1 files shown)


robertjamesmiller:
This pull request is negated by: https://github.com/sparc-request/sparc-request/pull/74 but maybe you accept our pull request first and then accept your pull request to make it easy for us to update our code base, thanks.

PR #80, opened 2015-05-11
Update to import of cores and services

robertjamesmiller:


amcates:
Can you explain your reasoning behind changing the report from using headers to using row position?

amcates:
Headers in my opinion make it easier to understand what the import is doing.

robertjamesmiller:
I agree that headers are easier to read and understand. Our end users were changing the header values which caused those data points to not be parsed.

amcates:
Sounds more like a user training issue.  Otherwise, I would be afraid that they wouldn't put the correct information in the correct columns.

robertjamesmiller:
The .csv file still has the header values but the import script is not relying on them to have an exact spelling.

amcates:
Ok, this will probably not be incorporated in to the OS repo unless we find more institutions who have the same issue.  We prefer the header route and it hasn't posed a problem at MUSC.

robertjamesmiller:
the old version has bugs in it (e.g., applying tags to services instead of cores) so it would be best to accept the changes and update after the fact. plus, it will need to be updated again once is_one_time_fee is moved from pricing_map to services.

lib/tasks/import_cores_and_services.rake
```diff
@@ -3,128 +3,151 @@ namespace :data do
...
                       # create core
-                      core = program.cores.new(:name => row['Core (this is optional)'],
-                                          :abbreviation => row['Core (this is optional)'],
+                      core = program.cores.new(:name => row[3],
+                                          :abbreviation => row[3],
                                           :order => 1,
                                         #  :organization_id => organization_id,
-                                          :is_available => (row['Hide from users? (Y/N)'] == 'N' ? true : false))
+                                          :is_available => (row[7] == 'N' ? true : false))
+                      core.tag_list.add("clinical work fulfillment") if row[8] == 'Y'
```
amcates: @robertjamesmiller Are these the tags you are referring to?

amcates: @robertjamesmiller Are these the tags you are referring to?

robertjamesmiller: yes, and there was a bug with how I was populating otf_unit_type.

PR #102, opened 2015-07-01
Otf go api

fordsar:


fordsar:
This code validates in both admin portal and in sparc-proper that the otf quantities are not greater than the max or less than the min set out in that service's profile in catalog manger.  app/views/service_calendars/_one_time_fees.html.haml -contains only some spacing changes from sublime, please ignore. If you have questions ask Johanna.

PR #103, opened 2015-07-01
allows removal of services which are not in fulfillment

fordsar:


fordsar:
This code accounts for allowing services which are not in fulfillment to be removed. For some reason the way we were validating the data attributes on the html object in the js was not working, so we changed it so they will always be booleans and therefore able to be evaluated in the js.

PR #112, opened 2015-07-07
Admin remember status

kyle-glick:


app/helpers/portal/service_requests_helper.rb
```diff
@@ -57,8 +57,8 @@ def populate_fulfillment_time fulfillment
     end
   end
 
-  def default_display workflow_state
-    workflow_state == 'submitted' ? '' : 'display:none;'
+  def default_display workflow_state, desired_status
```
amcates: Why is workflow_state even being passed in if all we are using is desired_status?  I don't think this is done properly.

jug2: Jason,

Can you jump on chat?

Thanks!

Andrew – to be clear – you don’t think this feature was done properly in the code. Correct?

PR #115, opened 2015-07-15
Linked quantities fixes

ctt3:
adds text error when trying to navigate past service calendar if linked quantities don't add up.


ctt3:
screen shot 2015-07-15 at 10 32 07 am

PR #124, opened 2015-07-29
fixing Add/Update Authorized User button

ying-pbrc:
To correct myself, I found the problem was not due to any missing local variable, rather when updating an existing user, the user_list was not being reloaded. 

Is there any reason why the Add Authorized User button always sitting there? Doesn’t it flow better where it was? I don't want to spend too much time to reinvent the engine and break anything else. I only fixed the issue by simply moving the .add-user div back to the user_details_right and updating its css to dark-blue-button.


robertjamesmiller:
Looks good to me, thanks for fixing.

robertjamesmiller:
We've deployed this fix to our dev and test servers and it is working well. Please merge this pull request, thanks.

PR #141, opened 2015-09-15
[#103318498] Change "Get a Quote" to "Cost Estimate"

goleksiak:


amcates:
Don't you need a migration to move requests in "get_a_quote" to "get_a_cost_estimate"

db/migrate/20150915155404_change_service_request_quote_status_to_get_cost_estimate.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeServiceRequestQuoteStatusToGetCostEstimate < ActiveRecord::Migration
+  def up
+    ServiceRequest.where(status: 'get_a_quote').update_all status: 'get_a_cost_estimate'
```
amcates: I think you need to do this on SubServiceRequest, AvailableStatus, and PastStatus.  They all have a status column

PR #142, opened 2015-09-15
Jjh cancel survey

jayhardee9:


app/controllers/surveyor_controller.rb
```diff
@@ -51,6 +51,12 @@ def update
     super
   end
 
+  def destroy
+    survey_ids = Survey.where(access_code: params[:survey_code]).pluck(:id)
+    ResponseSet.where(survey_id: survey_ids, access_code: params[:response_set_code]).first.destroy
```
amcates: Are you actually receiving survey_ids or just a survey_id?  Just wondering because if it could return multiple then you would want to call destroy on each.  Something like:

ResponseSet.where(survey_id: survey_ids, access_code: params[:response_set_code]).each(&:destroy)

PR #144, opened 2015-09-18
Added printer-friendly-request.css.scss

wtholt:
added stylesheet to include css in print preview for sub-service and consolidated service requests. included link to stylesheet in portal application layout


amcates:
Do we know how this affects printing on other portal pages?

PR #148, opened 2015-09-25
Removed stylesheet from portal app layout

wtholt:
removed print friendly stylesheet from portal app layout. placed stylesheet_link_tag at top of partials. It works for me locally, I don't know if there is a convention on how to add specific stylesheets to specific views


ryan-sci:
Did this already get pushed to master?  If so, no pull request is necessary.

PR #151, opened 2015-10-02
Wth routes cleanup

wtholt:
Updated ruby syntax. Ran the entire test suite multiple times. Encountered some failures (13-14 to be more exact), but since I'm not so familiar with the application and knowing that the test suite has some failures in it, these don't appear to be the result of changing the routes.rb file. Charlie showed me yesterday that he had worked on the routes.rb file in one of his branches. It might be best to consolidate his work and mine. 

Deleted destroy action in protocols controller since it is not being used. +1 for memory boost!

Added traceroute gem to assist in discovering unused routes.


Gemfile
```diff
@@ -49,6 +49,7 @@ gem 'simplecov', require: false, group: :test
 gem 'sinatra'
 gem 'surveyor'
 gem 'therubyracer', '0.10.2', :platforms => :ruby
+gem 'traceroute'
```
goleksiak: I think this should be in the development gem block as it is not needed in prodcution

db/schema.rb
```diff
@@ -134,13 +134,11 @@
     t.string   "comment"
     t.string   "remote_address"
     t.datetime "created_at"
-    t.string   "request_uuid"
```
goleksiak: Schema change without a migration?

Gemfile
wtholt: Agreed

PR #152, opened 2015-10-06
KG - Update Catalog Overlords

kyle-glick:


kyle-glick:
This should also probably be updated for new Kyle when we are able to.

PR #154, opened 2015-10-07
go jh jw archived protocols

jwiel86:


robertjamesmiller:
@jwiel86 It looks like the API tests need to be updated to include or ignore the new "archived" field:
SPARCCWF::APIv1 GET /v1/protocols.json request for :full records should respond with an array of protocols and their attributes
      Failure/Error: expect(parsed_body['protocols'].map(&:keys).flatten.uniq.sort).to eq(expected_attributes)
        
        expected: ["archived", "billing_business_manager_static_email",

PR #206, opened 2015-11-11
Needed to wrap document title text.

jleonardw9:


app/views/service_requests/_document_list.html.haml
```diff
@@ -27,9 +27,9 @@
     %tr.header{:class => provider_css}
       %th= t("documents.headers.institution")
       %th= t("documents.headers.type")
-      %th= t("documents.headers.title")
+      %th{style: 'width:150px'}= t("documents.headers.title")
```
amcates: Inline style, really?

PR #209, opened 2015-11-12
Additional details

robertjamesmiller:
We've completed version 1.0 of Additional Details and eagerly await your feedback. All of the tests for it can be run with: bundle exec rspec spec/models/additional_detail/ spec/controllers/additional_detail/

Please let us know if you have any questions, thanks.


robertjamesmiller:
I updated our pull request to resolve merging conflicts created by the recent acceptance of other pull requests.

robertjamesmiller:
I updated our pull request to resolve the merging conflicts created by the Rails 4 upgrade. We still need jQuery upgraded so that the navigation buttons on the "STEP 3: Shared Documents & Notes" page will work. This pull request is ready to be reviewed and tested.

amcates:
So, this pull request does not upgrade jQuery, correct?   Should it be added since it's a requirement of the pull request.  Otherwise, what version do you recommend and does that break the rest of the app?

robertjamesmiller:
This pull request only upgrades jQuery for the "STEP 3: Shared Documents & Notes" page. We recommend upgrading jQuery for the entire site to jQuery v2.1.4 , which will require updating all deprecated methods like this one: https://jquery.com/upgrade-guide/1.9/#live-removed

We identified that the following deprecated method:
$(".provider_program_core_save").live 'click', ->
should change to
$(document).on 'click',".provider_program_core_save", ->

That method is found in 20+ ".js.coffee" files. We have a commit that updates all references to that method but reverted it because we thought it would be best if it was done separate from this pull request since it affects so many pages that are unrelated to Additional Details and to limit the scope of this pull request. Are you all up for upgrading the deprecated methods in the 20+ ".js.coffee" files?

wtholt:
@robertjamesmiller I'm updating jQuery to v2.1.4 in a PR. We had two conflicting jQuery libraries that were causing problems with the default library from the Gemfile. One was in app/assets/javascripts and the other is coming from the Surveyor gem. I removed the one in app/assets and required v2.1.4 in Surveyor gem, but that will most definitely break things there.

I think we should rename the manifest JS files in each directory. For example, instead of 'application.js' in app/assets/javascripts/admin, we should rename to 'admin.js'. Just my opinion, what do you think?

robertjamesmiller:
@wtholt I'm open to the idea of renaming to admin.js. We set up the JS and CSS for additional details so that it would be isolated and not conflicting with the rest of the site. Since you're upgrading the rest of the site please feel free to adjust our JS/CSS configuration to be whatever is best for the site as a whole. For example, on the "STEP 3: Shared Documents & Notes" page, we're including all of the JS/CSS we need but that may be unnecessary after the JS/CSS is upgraded for the entire shopping cart flow. 

I recommend testing Additional Details as is to see that it is working first before tweaking any of our JS/CSS configuration. Notably, this file lists the JS/CSS dependencies for Additional Details: https://github.com/sparc-request/sparc-request/blob/master/bower-additional-details.json . Let me know if you have any questions.

jleonardw9:
@robertjamesmiller Is there something we have to configure for this? In Sparc Proper, when you get to the document's page you're stuck there. The styling looks weird as well.
screen shot 2015-12-17 at 1 48 09 pm

robertjamesmiller:
@jleonardw9 On the "Notes & Documents" page there are two things going on. First, our use of Bootstrap and AngularJS requires that jQuery be upgraded (we recommend the latest version 2.1.4); after which the navigation buttons will work. Please reference my comment on this pull request from seven days ago which links to https://jquery.com/upgrade-guide/1.9/#live-removed . In the interim, you'll need to click your browser's back button and then skip ahead to Step 4 (not satisfactory for production use). 

Second, the styling is off because we've customized/suppressed some of the existing sparc-request CSS to let Bootstrap handle font sizes, buttons, and spacing. See our home page here: https://i-cart.icts.uiowa.edu/ to get an idea. I'm not sure the best way to solve that issue because our CSS changes affect all of the pages on the service request flow. Please try referencing our application.css and let me know how things look for you: https://i-cart.icts.uiowa.edu/assets/uiowa/application-f1f66deaf11bcab8da5c9f14343f1b24.css

Another option would be for us to update the Notes & Documents page to function similarly to how we have Portal/Dashboard functioning such that we send users to a standalone page (that doesn't have CSS/JS conflicts) and then have them click back to the Notes & Documents page after they're done filling out additional details. Ideally, the service request flow would be updated to use Bootstrap Grid and Buttons and then our stuff would work without modification but I realize that might be too big of a project to tackle all at once.

jleonardw9:
Ah, ok cool. I'm going to have to pull this off of master for now, because we have things in the pipeline that need to make their way up to production. We have a branch that we are currently working on where we are cleaning up our js files and updating jQuery. Please re-submit this pull request to wth-ml-js-file-reorganization-and-upgrade.

robertjamesmiller:
Sounds good and will do. Our Travis CI build has identified a scenario in which some of our Additional Detail tests fail because of timezone differences. I'm going to fix that first and then create the new pull request, thanks.

PR #213, opened 2015-11-16
System satisfaction survey correctly included

ctt3:


robertjamesmiller:
This looks good. Thank you for your quick work on it.

app/assets/javascripts/review.js.coffee
```diff
@@ -21,57 +21,8 @@
 #= require navigation
 
 $(document).ready ->
-  #Survey
   survey_offered = false
```
amcates: Why did you leave these in this file?  I would think you could remove them.

app/views/service_requests/review.html.haml
```diff
@@ -19,7 +19,7 @@
 -# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 - if SYSTEM_SATISFACTION_SURVEY != false
-  = javascript_include_tag 'review'
```
amcates: Does this page have a 'Save As Draft' option?  If so, it needs the 'review' code as well so that it give the appropriate messages

ctt3: The review js is always included as it is required in the application js.

PR #235, opened 2015-12-10
Jl rails4 upgrade

jleonardw9:


robertjamesmiller:
This migration fails with Rails 4: MovePositionFromVisitToVisitGroup and I'll try to fix it.

PR #239, opened 2015-12-11
Upgrade to Ruby 2.1.5

robertjamesmiller:
Similar to sparc-fulfillment, it seems like we should configure this project such that all institutions are running the same version of Ruby, what do you think?

Within our Travis-CI build, upgrading Ruby from 1.9.3 to 2.1.5 did not change any RSpec test results but did cut the total run time in half!


amcates:
Robert, we didn't include .ruby-version and .ruby-gemset as files in the repo so that it wasn't forced but .ruby-version.example is already at 2.1.5 and that is what we are running.  I'm not opposed to including .ruby-version but it could potentially cause problems as new versions would force updates when they may not be required.

robertjamesmiller:
Andrew, what do you mean by "it could potentially cause problems as new versions would force updates when they may not be required"?

amcates:
I spoke with the team and they are good with the change.  I'm just used to apps saying stuff like "works with Ruby 1.9.3 or Ruby 2.1.5" and less of use "2.1.5" but I think in this case it's fine.  Old habits die hard

PR #240, opened 2015-12-11
Updated github references to use https instead of git

robertjamesmiller:
We needed this change to support our dev server's proxy.


robertjamesmiller:
FYI: I needed to update app/models/cover_letter_sanitizer.rb so that delayed_jobs would start on our dev server. The code looked like it was trying to allow "em" tags but the test: spec/models/cover_letter_sanitizer_spec.rb was asserting the allowance of "p" tags, so I updated the code so that the test continued to pass.

app/models/cover_letter_sanitizer.rb
```diff
@@ -18,6 +18,6 @@
 # INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-class CoverLetterSanitizer < HTML::WhiteListSanitizer
-  self.allowed_tags = HTML::WhiteListSanitizer.allowed_tags - %w(em)
+class CoverLetterSanitizer < Rails::Html::WhiteListSanitizer
+  self.allowed_tags = %w(p)
```
amcates: Robert, this doesn't have to do with switching git to https and shouldn't be in this pull request.

amcates:
I just noticed your comment.  I thought it was removing 'em' tags from allowed_tags.

robertjamesmiller:
Ah, I now I see, was the old version subtracting em from allowed tags? The p tag was the only one being tested so things should work fine. Do you want me to make any adjustments to this pull request?

amcates:
I'm wondering if we even need this anymore.  I believe Cover Letters can't be accessed in the new setup.  Might just remove the "self.allowed_tags" all together.

robertjamesmiller:
Gotcha, we still need the "github references to use https instead of git" so maybe you accept the pull request and then take out the cover letter references and dependencies? Or, do you want me to do that?

amcates:
We'll take care of it

PR #243, opened 2015-12-17
Sj rails 4 report fix

Stuart-Johnson:


app/reports/service_pricing_report.rb
```diff
@@ -135,7 +135,7 @@ def where args={}
 
     service_organization_ids = Organization.all.map(&:id) if service_organization_ids.compact.empty? # use all if none are selected
 
-    service_organizations = Organization.find_all_by_id(service_organization_ids)
+    service_organizations = Organization.find(service_organization_ids)
```
amcates: Does new 'find' method take a array?

PR #248, opened 2016-01-05
SubServiceRequest is 'Ready for Fulfillment' if protocol is funded    

robertjamesmiller:
A SubServiceRequest is 'Ready for Fulfillment' if its protocol's funding_status is funded.


robertjamesmiller:
We've decided to cancel this pull request because we want to be able to push "unfunded" studies to fulfillment.

PR #249, opened 2016-01-06
Default for is_available now set to true for organizations and services.

jleonardw9:


amcates:
Will this change existing Nulls to 1?

jleonardw9:
Yeah, I just tested it locally.

PR #278, opened 2016-01-28
Ml visit calendar move visit button bugs

marklohr:


app/models/arm.rb
```diff
@@ -252,6 +219,10 @@ def set_arm_edited_flag_on_subjects
   end
 
   def update_visit_group_day day, position, portal= false
+    # position = position.blank? ? visit_groups.last.position : position.to_i
```
wtholt: You can delete this commented out code

PR #301, opened 2016-02-23
Broken tests in master

chrisortman:
Hi, this isn't ready to merge yet, but wanted to create the PR now to make sure I'm not duplicating work.

I was trying to make sure tests were passing on master branch and have found some that are failing, so these are my fixes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gem 'activerecord-import'
 gem 'activerecord-session_store'
-gem 'acts_as_list', github: 'swanandp/acts_as_list'
+gem 'acts_as_list', github: 'chrisortman/acts_as_list'
 gem 'acts-as-taggable-on'
 gem 'audited-activerecord'
 gem 'axlsx'
--- a/app/models/project_role.rb
+++ b/app/models/project_role.rb
@@ -91,11 +91,25 @@ class ProjectRole < ActiveRecord::Base
   end
 
   def validate_one_primary_pi
-    if self.protocol.project_roles.where(role: "primary-pi").include?(self) && self.role != "primary-pi"
-      errors.add(:role, "- Protocols must have a Primary PI.")
+
+    if !protocol_has_primary_pi?
+      errors.add(:must, "include one Primary PI.")
+      return false
+    elsif self.role == 'primary-pi' && protocol_has_primary_pi?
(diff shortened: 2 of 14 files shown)


Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gem 'activerecord-import'
 gem 'activerecord-session_store'
-gem 'acts_as_list', github: 'swanandp/acts_as_list'
+gem 'acts_as_list', github: 'chrisortman/acts_as_list'
```
amcates: Chris, why are you switching this to your version over the official version?

chrisortman: I have a PR open to aal to fix a bug in position computation
https://github.com/swanandp/acts_as_list/pull/188

I'm hoping it will get merged before I'm done with this one

app/models/project_role.rb
```diff
@@ -91,8 +91,14 @@ def validate_other_selections
...
-    if self.protocol.project_roles.where(role: "primary-pi").include?(self) && self.role != "primary-pi"
-      errors.add(:role, "- Protocols must have a Primary PI.")
+
+    primary_pi = protocol.project_roles.where(role: 'primary-pi')
+
+    if primary_pi.empty?
+      errors.add(:must, "include one Primary PI.")
+      return false
+    elsif self.role == 'primary-pi' && !primary_pi.include?(self) && primary_pi.size > 0
+      errors.add(:role, "- This protocol already has a Primary PI.")
```
jayhardee9: Your changes to #validate_one_primary_pi allows users to remove Primary PI's from protocols when editing their associated users (protocols must have one Primary PI). This may be what you're looking for, fixing the project_role_spec:

```
def validate_one_primary_pi
    pi_roles = self.protocol.project_roles.where(role: "primary-pi")
    if pi_roles.empty? || pi_roles.include?(self) && self.role != "primary-pi"
      errors.add(:role, "- Protocols must have a Primary PI.")
      false
    else
      true
    end
  end
```

Gemfile
jleonardw9: Chris, I'm going to go ahead and close this one. Thanks for the work, we incorporated a lot of your suggestions into our test fixes, but we didn't see the need to update Acts as List for the test suite.

chrisortman:
Sounds good

PR #303, opened 2016-02-29
Jw epic validation bug

jwiel86:
Ran specs, diffed with master --> good to go.

### DIFF
--- a/spec/features/portal/admin/admin_edit_study_spec.rb
+++ b/spec/features/portal/admin/admin_edit_study_spec.rb
@@ -57,6 +57,13 @@ RSpec.describe 'editing a study', js: true do
       wait_for_javascript_to_finish
       expect(page).to have_content('1 error prohibited this study from being saved')
     end
+
+    it "should raise an error message if study's sponsor name is not filled out" do
+      fill_in 'study_sponsor_name', with: ''
+      click_button 'Save'
+      wait_for_javascript_to_finish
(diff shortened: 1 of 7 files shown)


spec/features/portal/admin/admin_edit_study_spec.rb
```diff
@@ -57,6 +57,13 @@
       wait_for_javascript_to_finish
       expect(page).to have_content('1 error prohibited this study from being saved')
     end
+
+    it "should raise an error message if study's sponsor name is not filled out" do
+      fill_in 'study_sponsor_name', with: ''
+      click_button 'Save'
+      wait_for_javascript_to_finish
+      expect(page).to have_content('1 error prohibited this study from being saved')
```
wtholt: Could we test this with a unit test (model spec perhaps) rather than an integration test (feature)?

jwiel86: Hey Will!
This was an odd error that was only happening in admin/portal. If sponsor name was not filled out it would count it as two errors and relay the same error message twice.  I asked J about it and he says that it would be too difficult to test in the model since it is an edge case and it wouldn't make sense to put it in the controller spec since we don't know what's causing the error to duplicate it in the first place.  He did say that he thought it was unnecessary to have that feature spec, so I will happily remove it! #operationkeepthespecsfixed

jleonardw9:
Merging this in. Please refactor 'Update Protocol Type' in the controller when time allows.

PR #311, opened 2016-03-15
Jl add status to consolidated

jleonardw9:
Branch name is misleading. This actually adds a line item's status to all calendar views, the export and consolidated view in portal, and the review page in sparc proper.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,19 +13,6 @@
 
 ActiveRecord::Schema.define(version: 20160229185056) do
 
-  create_table "additional_details", force: :cascade do |t|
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -13,6 +13,19 @@
 
 ActiveRecord::Schema.define(version: 20160229185056) do
 
+  create_table "additional_details", force: :cascade do |t|
```
amcates: Your db/schema.rb is ahead of the game

jleonardw9: Thanks for that, I'll get it fixed.

PR #317, opened 2016-03-15
KG - Bootstrap SSR Notification Counter

kyle-glick:


app/assets/javascripts/dashboard/notifications.js.coffee
```diff
@@ -78,11 +78,13 @@ $(document).ready ->
   $(document).on 'click', 'button.mark_as_read_unread',  ->
     selections = $('#notifications-table').bootstrapTable 'getSelections'
     notification_ids = selections.map( (hash, i) -> return hash['id'] )
+    sub_service_request_id = $(this).data('sub-service-request-id')
```
ctt3: Have you tested this on the notifications index page? (The red "Messages" button in the header) The index page has button.mark_as_read_unread, but is not specific to a ssr and does not have this data attribute.

kyle-glick: Oh yeah I think I was supposed to remove that line. I'll look into it

PR #320, opened 2016-03-16
Jw epic answer bug

jwiel86:


app/models/protocol.rb
```diff
@@ -129,6 +129,10 @@ def is_study?
   def virgin_project?
     selected_for_epic == nil
   end
+  # returns true if selected_for_epic is nil in the database
+  def selected_for_epic_is_nil_in_db?
```
amcates: What is this?  Seriously, is this a method.  What about just "selected_for_epic.nil?"
You already have the object.  Why are you re-querying?

app/models/protocol.rb
```diff
@@ -129,6 +129,10 @@ def is_study?
   def virgin_project?
     selected_for_epic == nil
```
amcates: and here you "== nil".  Consistancy please.  "selected_for_epic.nil?" is much cleaner

app/models/protocol.rb
amcates: If you are looking for changes that just happened or not you can use the rails helpers.  Look at these http://apidock.com/rails/ActiveRecord/Dirty

jwiel86: In sub_service_requests_controll#update_from_project_study_information, if the set of attributes passed in from the client is invalid for some reason and @protocol.update_attributes fails, we need to maintain the state of the @protocol record so the values the user input are maintained when the page is re-rendered. The value of @protocol.selected_for_epic is being set to true/false in memory though, rather than staying as nil, so the view is rendered incorrectly (displaying the epic box as readonly, rather than making it editable). My solution was therefore to just load the value of @protocol._selected_for_epic from the database instead of relying on the possibly (incorrectly) true/false value in memory. I guess this isn't ideal and I think I found a way around it using some built in ActiveRecord methods.

sparc-request: I wonder is selected_for_epic_was would help?  FYI, I was being overly harsh but we really should find a better way.

On Mar 16, 2016, at 6:46 PM, jwiel86 <notifications@github.com<mailto:notifications@github.com>> wrote:




In app/models/protocol.rb<https://github.com/sparc-request/sparc-request/pull/320#discussion_r56429211>:

> @@ -129,6 +129,10 @@ def is_study?
>    def virgin_project?
>      selected_for_epic == nil
>    end
> +  # returns true if selected_for_epic is nil in the database
> +  def selected_for_epic_is_nil_in_db?


In sub_service_requests_controll#update_from_project_study_information, if the set of attributes passed in from the client is invalid for some reason and @protocol.update_attributes fails, we need to maintain the state of the @protocol<https://github.com/protocol> record so the values the user input are maintained when the page is re-rendered. The value of @protocol<https://github.com/protocol>.selected_for_epic is being set to true/false in memory though, rather than staying as nil, so the view is rendered incorrectly (displaying the epic box as readonly, rather than making it editable). My solution was therefore to just load the value of @protocol._selected_for_epic from the database instead of relying on the possibly (incorrectly) true/false value in memory. I guess this isn't ideal and I think I found a way around it using some built in ActiveRecord methods.

PR #327, opened 2016-03-22
KG - Bootstrap Compose Notification

kyle-glick:


app/controllers/dashboard/notifications_controller.rb
```diff
@@ -41,9 +41,19 @@ def index
   end
 
   def new
-    @sub_service_request = SubServiceRequest.find(params[:sub_service_request_id])
-    @notification = @sub_service_request.notifications.new
-    @message = @notification.messages.new(to: params[:identity_id])
+    @sub_service_request_id = params[:sub_service_request_id]
+
+    if params[:identity_id]
+      @notification =
```
ctt3: This syntax is slick, but is harder to read than if you moved "@notifcation =" down to be in front of "@sub_service_request.notifications.new" and "Notification.new"

PR #329, opened 2016-03-23
Ml add hover effect to buttons

marklohr:


app/models/visit_group.rb
```diff
@@ -35,7 +35,6 @@ class VisitGroup < ActiveRecord::Base
   attr_accessible :window_before
   attr_accessible :window_after
   acts_as_list scope: :arm
-  scope :at_position, ->(position) { where(position: position) }
```
wtholt: Why did you remove this query?

marklohr: I meant to remove this on my previous story / branch (PR #278) which was already merged in.  It is not used anywhere (see first two commit messages)...   I probably should have done this straight to master branch.

config/locales/en.yml
```diff
@@ -105,14 +105,14 @@ en:
 
   # Bottom navigation
   bottom_navigation:
-    cancel: "Cancel"
-    continue: "Continue"
+    cancel: " ⟵ Cancel"
```
marklohr: Swapped out arrow .png files with unicode arrows.  Allows for better grouping during the hover state effect.  Loads faster than image file.

PR #334, opened 2016-03-29
KG - Bootstrap No Notifications to Self

kyle-glick:


app/views/dashboard/notifications/new.js.coffee
```diff
@@ -17,7 +17,11 @@
 # DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
 # INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
+<% if @errors.present? %> #Sending notification to self
+$("#modal_errors").html("<%= escape_javascript(render(partial: 'shared/modal_errors', locals: {errors: @errors})) %>")
```
wtholt: You could just do

     $('#modal_errors').html("<%= j render 'shared/modal_errors', errors: @errors %>')

app/controllers/dashboard/notifications_controller.rb
```diff
@@ -52,7 +52,12 @@ def new
         @notification = Notification.new
       end
 
-        @message = @notification.messages.new(to: params[:identity_id])
+      if Identity.find(params[:identity_id]) == current_user
```
jayhardee9: ActiveRecord delegates `==` to comparing the `:id` columns, so technically you could write `if params[:identity_id] == current_user.id` here instead, with the benefit of not hitting the database with `#find`. Also, in the event that `params[:identity_id]` refers to `current_user`, are you still wanting to run line 60?

PR #337, opened 2016-04-05
Jl jw add service status to admin

jleonardw9:


wtholt:
Only issue I see here is using new Ruby syntax. Other than that, looks good.

PR #349, opened 2016-04-14
Feature/cas

loganfreeman:
add support for omniauth_cas

need to add the following setting in application.yml to enable it:

```
use_cas: true
use_cas_only: false 
```

use_cas: flag to turn on and off omniauth_cas
use_cas_only: if false will use both database and omniauth_cas


app/controllers/identities/omniauth_callbacks_controller.rb
```diff
@@ -30,5 +30,19 @@ def shibboleth
       redirect_to new_identity_registration_url
     end
   end
-end
 
+  def cas
+    @identity = Identity.find_for_cas_oauth(request.env["omniauth.auth"], current_identity)
+
+    if @identity.persisted?
+      sign_in_and_redirect @identity, :event => :authentication
```
wtholt: We're working on better styling our Ruby code. Reference - https://github.com/bbatsov/ruby-style-guide

Here, use new Ruby syntax instead of hash rockets, e.g.

     event: :authentication

Parentheses for method arguments

     sign_in_and_redirect(@identity ...)

config/locales/en.yml
```diff
@@ -1124,10 +1124,10 @@ en:
   login_password: "Password:"
   remember_user: "Remember me:"
   signin:
-    button1: "MUSC Users Click Here"
```
jleonardw9: Please remove en.yml from your pull request. We need to keep our MUSC wording in place.

Gemfile
```diff
@@ -55,6 +55,8 @@ gem 'surveyor', :git => 'https://github.com/caboteria/surveyor.git', branch: 'ra
 gem 'therubyracer', '0.10.2', :platforms => :ruby
 gem 'uglifier', '>= 1.0.3'
 gem 'will_paginate'
+gem 'omniauth-cas'
```
jleonardw9: It looks like this gem is no longer maintained, the last contribution was 2 years ago. You may need to find another solution.
screen shot 2016-04-15 at 3 25 14 pm

app/controllers/identities/omniauth_callbacks_controller.rb
loganfreeman: Thanks

Gemfile
rkmadsen: As far as I can tell omni-cas is the only existing solution at this point. Many of the libraries that are used in SPARC are many years old. If this is the metric that code can be contributed we will be stuck in adding new solutions that use existing code that support the libraries that are used, such as omni-auth. Are you suggesting that we must write a new omni-auth plugin of our own and then get then contribute that?

rkmadsen: We'll can look at possibly taking over that gem or writing a new implementation but for now would it  be possible to include it as an option for CAS auth?

sparc-request: Yes, let me look at the gem in the morning. It could be that it's just fine.

On Apr 18, 2016, at 6:46 PM, rkmadsen <notifications@github.com<mailto:notifications@github.com>> wrote:




In Gemfile<https://github.com/sparc-request/sparc-request/pull/349#discussion_r60147597>:

> @@ -55,6 +55,8 @@ gem 'surveyor', :git => 'https://github.com/caboteria/surveyor.git', branch: 'ra
>  gem 'therubyracer', '0.10.2', :platforms => :ruby
>  gem 'uglifier', '>= 1.0.3'
>  gem 'will_paginate'
> +gem 'omniauth-cas'


We'll can look at possibly taking over that gem or writing a new implementation but for now would it be possible to include it as an option for CAS auth?

jleonardw9: I looked at the gem some more last night, and I think it looks good. So I think the only issues that need to be corrected are the Ruby styling and the customization of your own en.yml file. Check out this guide we wrote up to help you with your customizations: https://github.com/sparc-request/sparc-request/blob/master/CUSTOMIZE.md

loganfreeman:
Sorry, my fault. I will reopen it.

loganfreeman:
I will correct the styling and yml file and resend the request

Gemfile
rkmadsen: Great thanks for reviewing it and having patience with us! Sorry for the confusion on the pull requests, we're just getting started on sending changes your way.

loganfreeman:
I corrected the styling issue, removed the change to en.yml. 
I added a helper to deal with the issue of sign in button text using different institution name. 

Please review

loganfreeman:
just found there is a local_override. removed the helper, and just use key from local file.

app/models/identity.rb
```diff
@@ -56,25 +55,25 @@ class Identity < ActiveRecord::Base
...
-  has_many :catalog_manager_rights, :class_name => 'CatalogManager'
-  has_many :service_providers, :dependent => :destroy
-  has_many :notifications, :foreign_key => 'originator_id'
-  has_many :sent_messages, :class_name => 'Message', :foreign_key => 'from'
-  has_many :received_messages, :class_name => 'Message', :foreign_key => 'to'
-  has_many :user_notifications, :dependent => :destroy
-  has_many :received_toast_messages, :class_name => 'ToastMessage', :foreign_key => 'to', :dependent => :destroy
-  has_many :sent_toast_messages, :class_name => 'ToastMessage', :foreign_key => 'from', :dependent => :destroy
-  has_many :notes, :dependent => :destroy
+  has_many :approvals, dependent: :destroy
```
jwiel86: Thank you for updating this file to follow our style guide specifications!

config/routes.rb
```diff
@@ -19,21 +19,28 @@
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 SparcRails::Application.routes.draw do
```
jwiel86: Thanks again for following the style guide!  We appreciate it!

app/controllers/identities/omniauth_callbacks_controller.rb
```diff
@@ -20,15 +20,26 @@
...
+      session['devise.shibboleth_data'] = request.env['omniauth.auth']
       redirect_to new_identity_registration_url
     end
   end
-end
 
+  def cas
+    @identity = Identity.find_for_cas_oauth(request.env['omniauth.auth'], current_identity)
+
+    if !@identity.nil?
```
amcates: Shouldn't this be @identity.persisted?  Just like def shibboleth on line 25.

loganfreeman:
```ruby
  def self.find_for_cas_oauth(auth, _signed_in_resource = nil)
    Directory.search(auth.uid).first
  end
```

Directory.search will search and then return the first found. If it is not nil, then it is always persisted. Should check nil first. once nil checking passes, no need to check persisted. 

The logic here is that if a user successfully authenticated through CAS, then always succeed. unless this is not always true.

amcates:
@loganfreeman I have a couple issues with this pull request.  All of which I think can be easily resolved.  I feel like the refactoring should be done separately from the actual logic added.  It makes it much harder to determine exactly what has changed.  Second, why did you extract from obis_setup?   This isn't something that was required for your pull request.  I would prefer that pull requests be concise and easily explainable.  

With my comment about 'persisted?', the system creates new local identities once authenticated.  This is both to make the system faster (returning identities verses querying external systems) and it gives us a hard record of what the identity looked like.  

I think prior to future pull requests we should probably have a 10-15 conversation.  

Thanks,

Andrew

loganfreeman:
@amcates  for the obis_setup, its purpose is to load some constants. But the calling of the initializers has some order, such that the obis_setup initializer is not the first one to be called and the constants are not available for those called before obis_setup.
So I move the constants loading stuff to environement.rb.

loganfreeman:
@amcates I can make a new pull request without the obis_setup change. the devise initializer needs to know from application.yml what autheticattion provider to use. but it is called after obis_setup. So it needs read the application.yml again.

PR #350, opened 2016-04-14
Feature/assets

loganfreeman:
add support to override asset loading path.

just create a directory in Rails.root called themes. and put assets inside it. it will be the first to be loaded.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -535,7 +535,7 @@ en:
     mail_link: "Sign Up For the SPARC Request Listserv"
     mail_subject: "SPARC Request Listserv"
     link_one_text: "SPARC Request Blog"
-    link_one_url: "http://sparcrequestblog.com/"
+    link_one_url: "https://sparcrequestblog.com/"
     link_two_text: "SCTR"
     link_two_url: "http://sctr.musc.edu"
     link_three_text: "SPARC Request Home"
--- a/spec/support/webmock.rb
+++ b/spec/support/webmock.rb
@@ -3,7 +3,7 @@ WebMock.disable_net_connect!(allow_localhost: true)
 RSpec.configure do |config|
 
   config.before(:each) do
-    stub_request(:get, "http://www.sparcrequestblog.com/").
+    stub_request(:get, "https://www.sparcrequestblog.com/").
      to_return(status: 200, body: "")
   end
 end


wtholt:
You can easily add assets through vendor/assets then include them in manifest files. Why do it this way?

loganfreeman:
The CUSTOM_ASSET_PATH is a mess by itself.


I did a search on CUSTOM_ASSET_PATH.

and found:

      = link_to image_tag("#{CUSTOM_ASSET_PATH}about.png", class: 'button-hover'), "javascript:void(0);", class: "about_sparc_request"


      = image_tag "/assets/#{CUSTOM_ASSET_PATH}department_name_460_by_60.gif",class: 'header_logo_2', :alt => "department_logo"

some of them use relative path, others use absolute path.

Do you have to include CUSTOM_ASSET_PATH in all the places where custom images are required?

it cluttered the code. 

and about the search path:

the CUSTOM_ASSET_PATH directory hasn't been added to the search path.

So the above image tag  image_tag("#{CUSTOM_ASSET_PATH}about.png" will result in 404 not found since CUSTOM_ASSET_PATH is not even in the search path for assets.


[Search path](http://guides.rubyonrails.org/asset_pipeline.html)

```
2.2.1 Search Paths
When a file is referenced from a manifest or a helper, Sprockets searches the three default asset locations for it.

The default locations are: the images, javascripts and stylesheets directories under the app/assets folder, but these subdirectories are not special - any path under assets/* will be searched.
```

loganfreeman:
according to:
`~/.rvm/gems/ruby-2.1.5/gems/sprockets-rails-2.3.3/lib/sprockets/railtie.rb`


any directories under app/assets directory will be added to assets_path.
```ruby
  class Engine < Railtie
    # Skip defining append_assets_path on Rails <= 4.2
    unless initializers.find { |init| init.name == :append_assets_path }
      initializer :append_assets_path, :group => :all do |app|
        if paths["app/assets"].respond_to?(:existent_directories)
          app.config.assets.paths.unshift(*paths["vendor/assets"].existent_directories)
          app.config.assets.paths.unshift(*paths["lib/assets"].existent_directories)
          app.config.assets.paths.unshift(*paths["app/assets"].existent_directories)
        else
          app.config.assets.paths.unshift(*paths["vendor/assets"].paths.select { |d| File.directory?(d) })
          app.config.assets.paths.unshift(*paths["lib/assets"].paths.select { |d| File.directory?(d) })
          app.config.assets.paths.unshift(*paths["app/assets"].paths.select { |d| File.directory?(d) })
        end
      end
    end
  end
end
```

You don't even custom_asset_path.

All you need do is add a image_override directory under app/assets. then it will find it.

wtholt:
Sounds good to me

PR #351, opened 2016-04-14
Feature/bootstrap

loganfreeman:
add bootstrap-rails


wtholt:
We're moving away from Gemified assets and towards Bower for managing our assets.

loganfreeman:
Bower is good. I totally agree with moving away from Gemified assets

PR #353, opened 2016-04-14
Feature/user search

loganfreeman:
Do not create user when searching user, only create user when selected


app/controllers/identities_controller.rb
```diff
@@ -91,4 +91,30 @@ def disapprove_account
     @identity.update_attribute(:approved, true)
     Notifier.account_status_change(@identity, false).deliver unless @identity.email.blank?
   end
+
+  def search_or_create_identity
```
wtholt: This method looks pretty gnarly. Maybe extract into smaller private methods?

loganfreeman:
since route 'search/identities' is used as a list for the user search control, it should not create the user. It should return the users found from ldap or database.

once a user is selected from the user search control, it sends a post request to route '/identities/find_or_create' to find or create the identity using method Directory::search.

PR #373, opened 2016-05-05
SPARCRequest navigation changes

ying-pbrc:
Please review and let me know if you have any comments/suggestions.
Thanks,
Ying

### DIFF
--- a/app/views/shared/_navbar.html.haml
+++ b/app/views/shared/_navbar.html.haml
@@ -1,7 +1,20 @@
 %ul.navbar
-  - NAVBAR_LINKS.each do |name, path|
-    %li.dashboard
-      = link_to name.to_s, path, target: '_blank'
+  %li.dashboard
+    = link_to NAVBAR_BUTTONS['dashboard'], portal_root_path
+  - if  current_user
+    - if (current_user.is_super_user? || current_user.is_service_provider?)
+      %li.admin
+        = link_to NAVBAR_BUTTONS['admin'], portal_admin_path
+      - if current_user.is_super_user?
+        %li.reports
+          = link_to NAVBAR_BUTTONS['report'], reports_path
+    - if current_user.is_fulfillment_provider?
+      %li.fulfillment
+        = link_to NAVBAR_BUTTONS['fulfillment'], CLINICAL_WORK_FULFILLMENT_URL, target: '_blank'
+    - if (current_user.is_overlord? || current_user.is_catalog_manager?)
+      %li.catalog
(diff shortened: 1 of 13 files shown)


config/application.yml.example
```diff
@@ -58,11 +58,6 @@ development:
   system_satisfaction_survey: false
   editable_statuses: { 14: ['first_draft', 'draft', 'submitted', nil, 'get_a_cost_estimate', 'awaiting_pi_approval'],
                      48: ['draft', 'submitted'] }
```
jleonardw9: I think we need to leave the navbar links configurable in the yml file. Some institutions may not want a fulfillment link and the SPARCAdmin link will be going away when we go live with our bootstrap branch. It also makes it easy to point these different places for testing.

app/views/shared/_navbar.html.haml
```diff
@@ -1,7 +1,20 @@
 %ul.navbar
```
jleonardw9: See comment below.

config/locales/en.yml
```diff
@@ -35,7 +35,7 @@ en:
   feedback: "Feedback"
   issue_contact: "Please contact the SUCCESS Center at (843) 792-8300 or success@musc.edu for assistance with this process or with any questions you may have."
   portal_title: "SPARC Request Portal"
```
jleonardw9: Please do not make changes to this file, we need to maintain out own configuration. If you look in the readme, we show you how to set your own file up for localization.

ying-pbrc: I can undo the change. However, the original text is too long, it doesn't fit the button - see attached screenshot. Any other suggestions?
create an account

config/application.yml.example
ying-pbrc: I wasn't aware of the new Dashboard change when this pull request was submitted. With that in mind, let me rework on this.

config/locales/en.yml
jleonardw9: Sorry, you're right. I had the button turned off locally and didn't see the problem. Looks good!

config/application.yml.example
ying-pbrc: I moved the navigation links configurable from constant.yml to application.yml; modified the navbar partial and the index view in catalog_manager/catalog. I changed the name from navbar_links to nav_links as they are not limited to the navbar usage.

ying-pbrc:
After seeing the new dashboard change in v.1.7.0, I believe this pull request is no longer necessary.

PR #431, opened 2016-05-27
fix method which could divide by zero

amcates:


app/models/approved_subsidy.rb
```diff
@@ -45,7 +45,12 @@ def approved_percent_of_total
     # Calculates the percent of total_at_approval that is subsidized
     # (stored total - pi_contribution) / stored total then convert to percent
     total = total_at_approval
-    ((( total - pi_contribution ).to_f / total ) * 100.0 ).round(2)
+
+    if total.nil? || total == 0
```
wtholt: Nitpicky

     if total.nil? || total.zero?

PR #434, opened 2016-05-30
SAW-Hid the IND on hold fields 

SamWord:
Hid the IND on hold fields in the new/edit study and view study details section.


wtholt:
This is just commenting out code. Is this field going away permanently or do we need to come up with a way to programmatically show/hide this field?

jleonardw9:
This field can just be deleted. If we need to bring it back at some point, we have the history to look at.

PR #437, opened 2016-05-30
SAW-Added validations to line item visits subject count

SamWord:


app/views/dashboard/service_calendars/master_calendar/pppv/_pppv_calendar.html.haml
```diff
@@ -46,9 +46,13 @@
       title: 'Edit subject count',
       validate: function(val) {
         var n = ~~Number(val);
+        var max_subject_count = $(this).data('max-subject-count');
```
wtholt: Code looks fine to me. This file really needs to be refactored though. I know you are just working with what was already there, but having JS code inline with views defeats the purpose of MVC.

PR #438, opened 2016-05-31
KG - Locked SSR Bugs

kyle-glick:


app/helpers/dashboard/service_requests_helper.rb
```diff
@@ -21,10 +21,10 @@
 module Dashboard::ServiceRequestsHelper
 
   def modify_request_button_url(service_request)
-    if service_request.is_editable? # Request not locked
-      "/service_requests/#{service_request.id}/catalog?from_user_portal=true&edit_original=true&locked=false"
-    else # Request locked
+    if service_request.has_locked_ssr? # Request not locked
```
wtholt: We could make this a bit cleaner

    url = catalog_service_request_path(from_user_portal: true, edit_original: true, locked:  variable)

   if variable true, change the url, if false, change the url

app/models/service_request.rb
```diff
@@ -566,6 +560,16 @@ def is_editable?
     return true
   end
 
+  def has_locked_ssr?
+    sub_service_requests.each do |ssr|
+      if !ssr.can_be_edited?
+        return true
```
wtholt: I don't think you need to explicitly 'return' a boolean value

kyle-glick: I tried removing the returns, and the functionality did not work the same way. The return appears to be needed

amcates: Could you call 'map' and then 'any?'

kyle-glick: (sub_service_requests.map { |ssr| !ssr.can_be_edited }).any?

kyle-glick: I think we could, but by using the early return, we can potentially save time by not going through all SSRs. A single one which can't be edited means the SR should show, so we shouldn't have to continue iterating at that point

PR #441, opened 2016-05-31
Fixed logic for document errors. 1.7.0

jleonardw9:


app/controllers/service_requests_controller.rb
```diff
@@ -574,7 +574,7 @@ def document_save_update errors
     doc_type_other = params[:doc_type_other]
     upload_clicked = params[:upload_clicked]
 
-    if doc_type and process_ssr_organization_ids and (document or document_id)
+    if !doc_type.empty? && process_ssr_organization_ids && document
```
wtholt: Could we say

    doc_type.any?

jleonardw9: Good point. Making the change.

kyle-glick: Can't call .any? on Strings

jleonardw9: Right, and doc_type will either be an empty string or a string like 'Protocol', etc.

PR #475, opened 2016-06-08
WTH - SLA qty requested not saving

wtholt:
Fixed validation error on creating a line item. Was defaulting to 1,
because it could not find method parameter due to improper format.

checking for different formats when line items are created


kyle-glick:
@wtholt 
Broken spec!

PR #478, opened 2016-06-08
KG - Admin Protocol Filters to Radios (1.7.5)

kyle-glick:


app/controllers/dashboard/protocols_controller.rb
```diff
@@ -38,10 +38,10 @@ def index
     # if we are an admin we want to default to admin organizations
     if @admin
       @organizations = Dashboard::IdentityOrganizations.new(@user.id).admin_organizations_with_protocols
-      default_filter_params[:filtered_for_admin]       = @user.id.to_s
+      default_filter_params[:admin_filter] = "for_admin #{current_user.id}"
```
kyle-glick: We can pass the type of filter we want and the ID as a string to this scope. May also consider using array, but didn't to avoid using it in database protocol filters

app/models/protocol.rb
```diff
@@ -164,6 +163,16 @@ class Protocol < ActiveRecord::Base
       where.not(project_roles: { project_rights: 'none' })
   }
 
+  scope :admin_filter, -> (params) {
```
kyle-glick: This filter breaks apart the input string to determine which filter to use and the ID to pass

PR #498, opened 2016-06-14
JTM - Re-added function to protocol model (1.7.5)

BigChiefSmidgeums:


app/models/protocol.rb
```diff
@@ -460,35 +460,11 @@ def any_service_requests_to_display?
     return self.service_requests.detect { |sr| !['first_draft'].include?(sr.status) }
   end
 
-  def has_per_patient_per_visit? current_request, portal
+  def has_line_items_of_type? current_request, portal, one_time_fee = false
```
kyle-glick: For styling guidelines, def has_line_items_of_type?(current_request, portal, one_time_fee=false)

kyle-glick: Should probably also pass type as a string "otf" or "pppv" instead of boolean value one_time_fee

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
```diff
@@ -89,15 +89,15 @@
                 %h4.panel-title
                   = t(:dashboard)[:service_calendars][:consolidated_request][:services][:header]
               .panel-body
-                - if @protocol.has_per_patient_per_visit? @service_request, @portal
+                - if @protocol.has_line_items_of_type? @service_request, @portal
```
kyle-glick: Should be called with parentheses as per the rails style guide: @protocol.has_line_items_of_type(@service_request, @portal)

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
```diff
@@ -89,15 +89,15 @@
...
                   - @protocol.arms.each do |arm|
                     .arm_calendar_container.calendar-container{ data: { arm_id: arm.id } }
                       - next if arm.visit_groups.empty?
                       = render partial: '/dashboard/service_calendars/master_calendar/pppv/pppv_calendar', locals: { service_request: @service_request, arm: arm, page: @pages[arm.id], pages: @pages, review: false, merged: true, study_tracker: false }
-                - if @protocol.has_one_time_fees? @service_request, true
+                - if @protocol.has_line_items_of_type? @service_request, portal = true, one_time_fee = true
```
kyle-glick: Don't we already have access to @portal?

app/models/protocol.rb
Stuart-Johnson: I totally fail to see how that would enhance readability in the slightest. It's pretty simple: One time fee, or not. Also, switching over to strings instead of boolean would make the rest of the method less semantically consistent.

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
kyle-glick: Should we pass values directly or assign as done to enhance readability?

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
```diff
@@ -89,15 +89,15 @@
...
                   .calendar-container
                     = render partial: '/dashboard/service_calendars/master_calendar/otf/otf_calendar', locals: { service_request: @service_request, sub_service_request: nil, portal: true, merged: true, tab: @tab, review: false, study_tracker: false, protocol: @protocol }
-                - unless @protocol.has_per_patient_per_visit?(@service_request, true) || @protocol.has_one_time_fees?(@service_request, true)
+                - unless @protocol.has_line_items_of_type?(@service_request, portal = true) || @protocol.has_line_items_of_type?(@service_request, portal = true, one_time_fee = true)
```
kyle-glick: Don't we already have access to @portal?

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
Stuart-Johnson: We are assigning them this way, so that you can tell from here what the true value is doing, instead of having to dig up the method and check.

The portal value was already being passed into this method pre-refactor. The method is in the protocol model, which most certainly does not have access to @portal instance variable, so I'm not sure what you're saying.

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
Stuart-Johnson: Again, not sure what you're saying.

app/models/protocol.rb
kyle-glick: having "of_type" in the name implies that I should see a type in the parameters of a call to the function

Stuart-Johnson: You do see a type, but it's one of two types, by definition, that is boolean in nature: one, or the other. Hence, boolean value. This also follows a convention seen elsewhere in the code, where a method having to do with services/line_items has a parameter of one time fee or not, as a boolean.

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
kyle-glick: As in line 92, @portal was passed as the second parameter. Why not in this case and line 100?

Stuart-Johnson: AH! Not I see what you're saying. That was actually already like that, but you're right, I think. Looks like when this was originally written, they didn't realize that @portal was already a variable, and you could just pass that in, instead of "true."

app/views/dashboard/service_calendars/_view_full_calendar.html.haml
Stuart-Johnson: See above

PR #538, opened 2016-06-29
SAW - refactor catalog organizations forms

SamWord:
The forms for Institutions, Providers, Programs and Cores seemed similar enough to refactor into a single shared form. 
Also refactored the controllers to work better with the new form and wrote specs for each controller.
No specs for the shared form yet and did not include Services in the refactor because they are not a type of Organization.


app/controllers/catalog_manager/organizations_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+# SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+class CatalogManager::OrganizationsController < CatalogManager::AppController
+  layout false
+  respond_to :js, :html, :json
+
+  def create
+    @organization.build_subsidy_map()
```
jleonardw9: unless is an institution

PR #571, opened 2016-07-15
Jtm admin subsidy bug

BigChiefSmidgeums:


amcates:
When viewing a SSR, you can add/edit subsidy.  Does this still validate?

PR #583, opened 2016-07-22
SAW - dashboard history add changed by column

SamWord:
Added changed_by to past statuses, moved update_past_status to controllers instead of sub_service_request model because it now requires current_user, wrote specs to ensure a past_status is being created where it is called.
Need to add spec to sub_service_requests/put_update_spec once it has been merged in.

Did not get previous status changes from the audit trail - that will be extracted into a separate story.


amcates:
Fix conflicts and re-review

app/controllers/service_requests_controller.rb
```diff
@@ -386,6 +388,7 @@ def add_service
           ssr.update_attribute :status, 'first_draft'
         elsif ssr.status.nil? || (ssr.can_be_edited? && ssr_has_changed?(@service_request, ssr) && (ssr.status != 'complete'))
           ssr.update_attribute :status, 'draft'
+          ssr.update_past_status(current_user) unless ssr.status.nil?
```
amcates: need to save off past status and then check against it

app/models/past_status.rb
```diff
@@ -23,9 +23,12 @@ class PastStatus < ActiveRecord::Base
 
   belongs_to :sub_service_request
 
+  belongs_to :changer, class_name: 'Identity', foreign_key: 'changed_by'
+
   attr_accessible :sub_service_request_id
   attr_accessible :status
   attr_accessible :date
+  attr_accessible :changed_by
```
amcates: Should be changed_by_id

PR #614, opened 2016-08-05
Jl zero percent subsidy validation

jleonardw9:


app/controllers/subsidies_controller.rb
```diff
@@ -23,7 +23,8 @@ class SubsidiesController < ApplicationController
 
   def create
     @sub_service_request = SubServiceRequest.find params[:subsidy][:sub_service_request_id]
-    @subsidy = PendingSubsidy.create(sub_service_request_id: @sub_service_request.id, percent_subsidy: 0)
+    @subsidy = PendingSubsidy.new(sub_service_request_id: @sub_service_request.id, percent_subsidy: 0)
+    @subsidy.save(validate: false)
```
wtholt: Why do we not validate here?

wtholt:
Looks good to me. See question about validation though.

app/controllers/subsidies_controller.rb
jleonardw9: Because the validation for zero percent subsidy is done through javascript in sparc proper. It's being done through model validations in dashboard, so I still need it to be in the model.

wtholt: 💯

PR #615, opened 2016-08-08
SAW - Add SSR Owner Filter on Dashboard

SamWord:
Filters protocols on the dashboard by Service Provider


app/models/identity.rb
```diff
@@ -125,6 +125,10 @@ def full_name
     "#{first_name.try(:humanize)} #{last_name.try(:humanize)}".lstrip.rstrip
   end
```
jleonardw9: Instead of adding another method on identity, you can do:
 full_name.split(' ').reverse.join(', ')

PR #627, opened 2016-08-15
Jw requester fix

jwiel86:


app/views/notifier/_welcome_for_submitted_status.html.haml
```diff
@@ -0,0 +1,19 @@
+- welcome_text = I18n.t('notifier.welcome_submit')
```
jayhardee9: This could be a little cleaner if the inline styles were moved into a separate stylesheet.

PR #637, opened 2016-08-18
Jjh ssr modal display

jayhardee9:


app/assets/stylesheets/dashboard/service_calendar.sass
```diff
@@ -1,3 +1,14 @@
+.user-view-ssr-modal
+  width: 95%
+
+.visit-before-container, .visit-day-container, .visit-after-container
```
kyle-glick: I think we should leave the class on the elements. It's a more verbose system than assigning it in the CSS. Thoughts?

jayhardee9: Sorry, I thought I responded to this yesterday. Anyways, I think it's a good idea to start moving Bootstrap classes into the stylesheets like this. After all, the view files should be devoted to markup, not styling.

Stuart-Johnson: If I understand Kyle correctly, I'm with him on this one. As far as I can tell, you are removing bootstrap classes from the views, and replacing them with your own custom classes, only to then import the styling mixin in the css, and assign it to your new custom selector. This will make for oversized custom css files eventually, and make maintaining styles across the site a nightmare. One of the benefits of using bootstrap in the first place is that you have a universal set of classes to use in your markup, and can avoid constantly creating your own (which then almost always leads to duplication, and unintended styling bugs).

jayhardee9: That's exactly what I'm doing, but I suppose it really is a game of whack-a-mole w.r.t. to styling and views/stylesheets. If you guys want the moles in the view, then I'll put them back.

PR #643, opened 2016-08-19
Sj modal closing changes

Stuart-Johnson:


app/assets/javascripts/dashboard/notes.js.coffee
```diff
@@ -42,4 +42,15 @@ $ ->
       type: 'GET'
       url: '/dashboard/notes/new'
       data: data
+
+  $(document).on 'click', 'button.notes.cancel',  ->
```
kyle-glick: Talked to Stuart about this, this does the same thing as the notes list button handler, should they be combined? Potentially we might see different functionality in the future for cancel, but right now they're the same

PR #648, opened 2016-08-23
Jjh protocol model refactor

jayhardee9:


app/models/protocol.rb
```diff
@@ -269,32 +276,20 @@ def validate_proxy_rights
...
     billing_business_manager_static_email.blank? ?  billing_managers.map(&:email).try(:join, ', ') : billing_business_manager_static_email
   end
 
-  def coordinators
-    project_roles.select{|pr| pr.role == 'research-assistant-coordinator'}.map(&:identity)
-  end
-
   def coordinator_emails
-    coordinators.map(&:email).try(:join, ', ')
+    coordinators.pluck(:email).try(:join, ', ')
```
kyle-glick: Is the join needed here? An empty pluck should return [] which can be joined to ''

app/models/protocol.rb
```diff
@@ -306,27 +301,20 @@ def primary_pi_exists
...
-  def role_for identity
-    project_roles.detect{|pr| pr.identity_id == identity.id}.try(:role)
+  def role_for(identity)
+    project_roles.find_by(identity_id: identity.id).try(:role)
   end
 
-  def role_other_for identity
-    project_roles.detect{|pr| pr.identity_id == identity.id}.try(:role)
+  def role_other_for(identity)
+    role_for(identity)
```
kyle-glick: This seems redundant? There's no separate utility from role_for(identity)

app/models/protocol.rb
```diff
@@ -306,27 +301,20 @@ def primary_pi_exists
...
   end
 
-  def subspecialty_for identity
+  def subspecialty_for(identity)
```
kyle-glick: Should this just be removed and replaced with identity.subspecialty?

app/models/protocol.rb
```diff
@@ -306,27 +301,20 @@ def primary_pi_exists
...
     identity.subspecialty
   end
 
   def all_child_sub_service_requests
```
kyle-glick: Should this be replaced with protocol.sub_service_requests? Redundant method

app/models/protocol.rb
```diff
@@ -443,31 +429,20 @@ def create_arm(args)
...
 
   def find_sub_service_request_with_ctrc(service_request)
-    service_request.sub_service_requests.each do |ssr|
-      if ssr.ctrc?
-        return ssr.ssr_id
-      end
-    end
-
-    return nil
+    service_request.sub_service_requests.find(&:ctrc?).try(:ssr_id)
```
kyle-glick: sub_service_requests.find(&:ctrc?).try(:ssr_id)

app/models/protocol.rb
```diff
@@ -478,28 +453,22 @@ def has_line_items_of_type?(current_request, portal, type)
...
 
-  def direct_cost_total service_request
-    total = 0
-    self.service_requests.each do |sr|
-      next if ['first_draft'].include?(sr.status) && sr != service_request
-      total += sr.direct_cost_total
-    end
-    return total
+  def direct_cost_total(service_request)
+    service_requests.where('status != ? OR id = ?', 'first_draft', service_request.id).
```
kyle-glick: service_request.where(id: service_request.id).where.not(status: 'first_draft')

app/models/protocol.rb
```diff
@@ -478,28 +453,22 @@ def has_line_items_of_type?(current_request, portal, type)
...
 
-  def indirect_cost_total service_request
-    total = 0
+  def indirect_cost_total(service_request)
     if USE_INDIRECT_COST
-      self.service_requests.each do |sr|
-        next if ['first_draft', 'draft'].include?(sr.status) && sr != service_request
-        total += sr.indirect_cost_total
-      end
+      service_requests.where('(status != ? AND status != ?) OR id = ?', 'first_draft', 'draft', service_request.id).
```
kyle-glick: service_requests.where(id: service_request.id).where.not(status: ['first_draft', 'draft'])

app/models/protocol.rb
jayhardee9: I think you're right.

app/models/protocol.rb
jayhardee9: It is, but I'm preserving the Protocol model's interface in this pull request. Just wanting to make memory usage, speed and style improvements. Once this PR is vetted however, we should totally rip out redundant methods and the like.

app/models/protocol.rb
jayhardee9: Need to scope through the service_request param.

PR #657, opened 2016-09-01
Jw dashboard status bug

jwiel86:
When sorting on dashboard page, params for scopes with_status, with_organization, and with_owner were being passed in as type String rather than type Array.  Also, with_owner was not being hit when sorting on dashboard.  Specs were written for each scope to test both types of passed in params.


app/models/protocol.rb
```diff
@@ -193,23 +193,26 @@ class Protocol < ActiveRecord::Base
 
   scope :with_status, -> (status) {
     # returns protocols with ssrs in status
-    return nil if status.reject!(&:blank?) == []
+    status = status.split.flatten.reject(&:blank?)
```
kyle-glick: Is this not passed as individual statuses?

app/models/protocol.rb
```diff
@@ -193,23 +193,26 @@ class Protocol < ActiveRecord::Base
 
   scope :with_status, -> (status) {
```
kyle-glick: Should rename 'status' accordingly. Perhaps statuses?

app/models/protocol.rb
```diff
@@ -193,23 +193,26 @@ class Protocol < ActiveRecord::Base
...
     # returns protocols with ssrs in status
-    return nil if status.reject!(&:blank?) == []
+    status = status.split.flatten.reject(&:blank?)
+    return nil if status.empty?
     joins(:sub_service_requests).
-    where(sub_service_requests: { status: status }).distinct
+      where(sub_service_requests: { status: status }).distinct
   }
 
   scope :with_organization, -> (org_id) {
```
kyle-glick: Should rename 'org_id' accordingly. Perhaps org_ids?

app/models/protocol.rb
```diff
@@ -193,23 +193,26 @@ class Protocol < ActiveRecord::Base
...
     # returns protocols with ssrs in org_id
-    return nil if org_id.reject!(&:blank?) == []
+    org_id = org_id.split.flatten.reject(&:blank?)
+    return nil if org_id.empty?
     joins(:sub_service_requests).
-    where(sub_service_requests: { organization_id: org_id }).distinct
+      where(sub_service_requests: { organization_id: org_id }).distinct
   }
 
   scope :with_owner, -> (owner_id) {
```
kyle-glick: Should rename 'owner_id' accordingly. Perhaps owner_ids?

spec/models/protocol_spec.rb
```diff
@@ -177,4 +177,255 @@
       expect{ study.push_to_epic(EPIC_INTERFACE) }.to change(EpicQueueRecord, :count).by(1)
     end
   end
+
+  describe "#with_status" do
```
kyle-glick: Recommendation: Break these out into individual specs. Testing 1 method/scope per spec file is easier to read. We do this for some models and most controllers.

PR #660, opened 2016-09-01
KG - Adding Sorting Span Spacing For Spec Fixes

kyle-glick:
**NOTE** Because the span is float: right this doesn't affect the position in the table at all. This simply gives some space for capybara to be able to click the element


jayhardee9:
Where is this spec breaking? Master is passing.

PR #662, opened 2016-09-08
Kg ssr submitted at fixes

Stuart-Johnson:


ying-pbrc:
don't forget to add submitted_at to the white list of the sub_service_request model attributes

ying-pbrc:
Also, in  db/migrate/20160908144020_add_submitted_at_to_sub_service_request.rb:
 
 +    SubServiceRequest.all.each do |ssr|
 +      past_status = ssr.past_statuses.order('date').last
 +
 +      unless past_status.nil?
 +        ssr.update_attribute(:submitted_at, past_status.date)
 +      end
 +    end

In each loop, 1) we probably should only considered those past statuses with a draft status, 2) only set ssr.submitted_at value if the status is not draft:

   past_status = ssr.pass_statuses.where(status: 'draft').order('date').last 

   unless past_status.nil? || ssr.status == 'draft'

PR #678, opened 2016-09-22
KG - Removing Unintentional Service Calendar Styles

kyle-glick:


kyle-glick:
This PR originally stemmed from fixing the new navy-blue coloring all over the dashboard, but I noticed that the same styling which was added back several days ago was causing this issue and the alignment issues in the visit headers.

PR #685, opened 2016-09-26
adding where clause to submitted_at migration for ssrs [#130020335]

Stuart-Johnson:


Stuart-Johnson:
FYI, when deployed, data needs to be refreshed, since this changes a migration.

PR #686, opened 2016-09-27
Jl calendar day bug

jleonardw9:


app/helpers/application_helper.rb
```diff
@@ -216,7 +216,7 @@ def visits_select_options(arm, pages)
       (beginning_visit..ending_visit).each do |visit_number|
         visit_group = arm.visit_groups[visit_number - 1]
 
-        if visit_group.day.present?
+        unless visit_group.day.nil?
```
jayhardee9: `if visit_group.day` would be equivalent, and a little more understandable, in my opinion.

wtholt: I'm fine either way. I think I prefer it the way it is. Both ways make sense to me.

jleonardw9: I accidentally changed that when I was experimenting. I'll change it back.

jleonardw9: Reverted back to original.

app/models/arm.rb
```diff
@@ -48,6 +48,14 @@ class Arm < ActiveRecord::Base
   validates :visit_count, numericality: { greater_than: 0 }
   validates :subject_count, numericality: { greater_than: 0 }
 
+  validate do |arm|
```
kyle-glick: Is there a better way of doing this? This seems too hard-coded

jleonardw9: I agree, but I couldn't figure out another way to make the validations chain back down to the visit groups. Let's think on it and try to refactor down the road.

PR #688, opened 2016-09-27
KG - Fixing Epic Question Initial DIsplay Logic (1.9.0)

kyle-glick:


app/assets/javascripts/dashboard/protocol_form.js.coffee
```diff
@@ -173,7 +173,7 @@ $(document).ready ->
 
   ###END PUBLISH IN EPIC BUTTON STATES###
 
-  if $("input[name='protocol[selected_for_epic]']").val() == 'true'
+  if $("input[name='protocol[selected_for_epic]'][value='true']").parent().hasClass('active')
```
jayhardee9: I think this would be a little more clearer: `if $(".active > input[name='protocol[selected_for_epic]'][value='true']")`. It's fine otherwise though, really.

Stuart-Johnson: Same, I'd much rather see CSS selectors used for this, instead another javascript selector to go to the parent and look at it's class. I think Jay's suggestion would also technically be a tiny bit faster to execute.

Stuart-Johnson: Same, I'd much rather see CSS selectors used for this, instead another javascript selector to go to the parent and look at it's class. I think Jay's suggestion would also technically be a tiny bit faster to execute.

kyle-glick: Agreed

kyle-glick: Tested this and didn't work. I'm going to try if $(".active > input[name='protocol[selected_for_epic]']").val() == 'true'

Stuart-Johnson: You know, I'm not sure the '>' is needed here... Might simplify to remove it. But I might be missing something.

Stuart-Johnson: Never mind, I didn't see the last commit when I wrote that. Disregard...

jleonardw9:
@kyle-glick I'm going to close this. Honna has all this fixed on a more recent branch

PR #703, opened 2016-10-18
Improve services lists load time on home page.

jayhardee9:
For example, pre-loading PricingMaps reduced loading MUHA > HB > Lab. services from 27 sec to 12 sec (still not great). And 3490 queries to 5 which is alright.


jayhardee9:
Already done in the bootstrap branch.

PR #706, opened 2016-10-19
WTH - Submission Form location and visibility adjustment

wtholt:
I moved the Additional Detail Submissions table underneath where Service
Requests are currently. Also, if there are no submissions for that
protocol, hide the submissions table.

I associated submissions to a Protocol and Line Item. That way I could
access more data points than the way it was set up
backend-wise. [#132617999]

### DIFF
--- a/app/controllers/additional_details/submissions_controller.rb
+++ b/app/controllers/additional_details/submissions_controller.rb
@@ -38,8 +38,10 @@ class AdditionalDetails::SubmissionsController < ApplicationController
     @service = Service.find(params[:service_id])
     @questionnaire = @service.questionnaires.active.first
     @submission = Submission.new(submission_params)
-    @protocol = Protocol.find(params[:protocol_id])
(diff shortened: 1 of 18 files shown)


app/controllers/additional_details/submissions_controller.rb
```diff
@@ -38,8 +38,10 @@ def create
     @service = Service.find(params[:service_id])
     @questionnaire = @service.questionnaires.active.first
     @submission = Submission.new(submission_params)
-    @protocol = Protocol.find(params[:protocol_id])
-    @service_request = ServiceRequest.find(params[:sr_id])
+    @protocol = Protocol.find(submission_params[:protocol_id])
+    @submissions = @protocol.submissions
+    @line_item = LineItem.find(submission_params[:line_item_id])
+    @service_request = ServiceRequest.find(@line_item.service_request_id)
```
jayhardee9: `@service_request = @line_item.service_request` might be clearer.

PR #711, opened 2016-10-24
Jw delete services email fix

jwiel86:


app/models/service_request.rb
```diff
@@ -608,6 +608,16 @@ def ssrs_associated_with_service_provider (service_provider)
...
   end
 
+  def ssrs_to_be_displayed_in_email(service_provider, audit_report, ssr_destroyed, ssr_id)
+    if ssr_destroyed
+      ssr = SubServiceRequest.find(ssr_id)
+      ssrs_to_be_displayed = [ssr] if service_provider.identity.is_service_provider?(ssr)
+    else
+      ssrs_to_be_displayed = self.ssrs_associated_with_service_provider(service_provider)
+    end
+    return ssrs_to_be_displayed
```
wtholt: I don't think you need to explicitly 'return' here on line 618, I think you could just say the variable name.

PR #720, opened 2016-10-28
Allow emails to be turned off per deployment

chrisortman:
We needed to be able to disable sending of emails to real users in our test and staging environments, but relying on a check for the Rails environment is problematic and was duplicated in several places in the code.

This change removes the duplication and contains the logic in a single place. It also moves the setting to application.yml.

This setting is intentionally off by default so that someone couldn't accidentally deploy to testing and have it start sending emails on them.
## 

I created this branch based on the release-1.8.0 tag thinking that it would then not pick up any changes that were not ready to be released. But it looks like the tags get created off musc_production but musc_production has stuff that's not in master?

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -53,8 +53,7 @@ class UserMailer < ActionMailer::Base
   private
 
   def send_message subject, is_service_provider='false', ssr_id=''
-    email = Rails.env == 'production' ? @send_to.email : DEFAULT_MAIL_TO
-    subject = Rails.env == 'production' ? subject : "[#{Rails.env.capitalize} - EMAIL TO #{@send_to.email}] #{subject}"
+    email = @send_to.email
     @is_service_provider = is_service_provider
     @ssr_id = ssr_id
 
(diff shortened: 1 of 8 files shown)


wtholt:
Please rebase this branch to our updated master. We just incorporated a lot of new changes into our current master branch. Fix the conflicts accordingly.

app/mailers/user_mailer.rb
```diff
@@ -51,8 +51,7 @@ def notification_received(user, ssr)
   private
 
   def send_message subject, is_service_provider='false', ssr_id=''
-    email = Rails.env == 'production' ? @send_to.email : DEFAULT_MAIL_TO
-    subject = Rails.env == 'production' ? subject : "[#{Rails.env.capitalize} - EMAIL TO #{@send_to.email}] #{subject}"
+    email = @send_to.email
```
kyle-glick: subject is missing here now. Is this intentional?

chrisortman: Yes intentional, subject is passed straight through to mail.

Otherwise it would have been `subject = subject` I think

chrisortman:
Rebase completed.

Please see also the lines I've marked with ##Review they didn't look correct in the original code to me.

jleonardw9:
Hey Chris. Can you look into the test failures? We'll get it merged in as soon as it's passing on Travis.

PR #721, opened 2016-11-01
Jjh back to top

jayhardee9:


config/locales/en.yml
```diff
@@ -204,6 +204,12 @@ en:
       header: "Study Cost Totals"
       grand_total: "Grand Total"
 
+  ###########
+  # CATALOG #
+  ###########
+  catalog:
```
kyle-glick: Can you move this to proper.en.yml? There's a section for catalog already in there.

PR #725, opened 2016-11-01
KG - Configurable Emails / Updated Emails

kyle-glick:


chrisortman:
I'm question if the list serve email address should change or be configurable. That seems like a different thing that the contact email address.

I'd also say that any time new configuration values are added like this they should provide defaults so that installing an update doesn't break an existing system.

config/initializers/obis_setup.rb
```diff
@@ -25,6 +25,9 @@
   ADMIN_MAIL_TO                             = application_config['admin_mail_to']
   EPIC_RIGHTS_MAIL_TO                       = application_config['approve_epic_rights_mail_to']
   FEEDBACK_MAIL_TO                          = application_config['feedback_mail_to']
+  CONTACT_US_MAIL_TO                        = application_config['contact_us_mail_to']
```
jayhardee9: Riffing of off Chris's comment, will the app break if these aren't defined in application.yml?

PR #730, opened 2016-11-04
Jw debugging cart remove add actions

jwiel86:


app/views/service_requests/add_service.js.coffee
```diff
@@ -17,4 +17,4 @@
 # DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
 # INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-$('.cart').replaceWith("<%= escape_javascript(render( 'service_requests/right_navigation/cart', service_request: @service_request, sub_service_request: @sub_service_request, sub_service_requests: @sub_service_requests, line_items_count: @line_items_count, allow_delete: true )) %>")
```
wtholt: 👍 - I think html() is the right choice here (http://stackoverflow.com/questions/730916/whats-the-difference-between-jquerys-replacewith-and-html)

app/controllers/service_requests_controller.rb
```diff
@@ -571,7 +572,7 @@ def send_individual_service_provider_notification(service_request, sub_service_r
     line_items = []
     @service_request.sub_service_requests.each do |ssr|
       if service_provider.identity.is_service_provider?(ssr)
-        line_items << SubServiceRequest.find(ssr).line_items
+        line_items << SubServiceRequest.find(ssr.id).line_items
```
kyle-glick: Why are we finding the SubServiceRequest when we already have it from the .each iteration? Can't we do line_items << ssr.line_items?

app/assets/javascripts/cart.js.coffee
```diff
@@ -45,6 +42,8 @@ $(document).ready ->
       $('#modal_place').modal('show')
       $('#modal_place .yes-button').data('srid', srid)
       $('#modal_place .yes-button').data('service-id', id)
+      $('#modal_place .yes-button').on 'click', (e) ->
```
kyle-glick: Is this a local handler? I noticed that the problem with the original code was that as you would remove/add more services, the handlers would add up and multiple clicks would be registered when you click the 'Yes' button. @jwiel86 Can you confirm?

If this does work, we should remove the two above lines setting data attributes on the yes button because they won't be needed anymore

jwiel86: @kyle-glick When I was debugging this file, the variables set after $(document).on 'click', '.remove-service', -> were not being reset.  So if you deleted one line_item, everything would work fine.  However, if you deleted a second one.  It would try to execute the code with the previous line item id and then it would also execute the code for the line item you just removed (I think that is what you were describing above- multiple clicks). Changing it to this selector on click format solved the problem for me.

app/controllers/service_requests_controller.rb
jwiel86: @kyle-glick I was fixing a deprecation warning here, but yes!  Good catch.

app/assets/javascripts/cart.js.coffee
jayhardee9: Pretty sure this onClick handler can only run once. Whole modal + .yes-button re-rendered on line 41.

PR #731, opened 2016-11-04
fixing save protocol filter error

ying-pbrc:
Receiving the following errors:

ActiveModel::MassAssignmentSecurity::Error (Can't mass-assign protected attributes for ProtocolFilter: sorted_by):
app/controllers/dashboard/protocol_filters_controller.rb:25:in `new'

### DIFF
--- a/app/models/protocol_filter.rb
+++ b/app/models/protocol_filter.rb
@@ -35,6 +35,7 @@ class ProtocolFilter < ActiveRecord::Base
   attr_accessible :with_organization
   attr_accessible :with_status
   attr_accessible :with_owner
+  attr_accessible :sorted_by
 
   scope :latest_for_user, -> (identity_id, limit) {
     where(identity_id: identity_id).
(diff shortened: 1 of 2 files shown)


app/models/protocol_filter.rb
```diff
@@ -35,6 +35,7 @@ class ProtocolFilter < ActiveRecord::Base
   attr_accessible :with_organization
   attr_accessible :with_status
   attr_accessible :with_owner
+  attr_accessible :sorted_by
```
kyle-glick: This isn't actually an attribute on the ProtocolFilter model anymore. We changed the design to do sorting via custom Javascript that calls an AJAX request for Dashboard::ProtocolsController#index.js and it passes the "sorted_by" attribute. I think what we can do instead is in Dashboard::ProtocolFiltersController#new we should reject that attribute from params[:filterrific]

Maybe something like this:
@protocol_filter = @user.protocol_filters.new(params[:filterrific].delete(:sorted_by)

PR #732, opened 2016-11-07
auto fix the calendar when it's not correct or missing data

amcates:


jleonardw9:
We're going to go ahead and close this one for now until it can be investigated and discussed further.

PR #740, opened 2016-11-15
Saw - Refactor update_past_status

SamWord:
update_past_status is now called after saving an SSR, and changed_by_id is set by getting the user who updated the status from the audit trail.
Because changed_by_id is set using the audit trail, a separate feature spec is used to test that the correct identity_id is saved to a past status.


app/models/sub_service_request.rb
```diff
@@ -368,13 +368,24 @@ def arms_editable?
...
+    #candidates.unshift("draft") if self.can_be_edited?
+    candidates << "ctrc review" if self.ctrc?
+    candidates << "ctrc approved" if self.ctrc?
+    candidates << "awaiting pi approval"
+    candidates << "on hold"
+
+    candidates
+  end
+
+  def update_past_status
```
jayhardee9: I think there could be race between the Audit being committed to the db and invocation of this after_save callback. Not sure if the update and the Audit creation are bundled together in a single db transaction. I would implement an `after_audit` instance method. [See this](https://github.com/collectiveidea/audited/blob/master/lib/audited/auditor.rb#L62)

app/models/sub_service_request.rb
```diff
@@ -368,13 +368,24 @@ def arms_editable?
...
 
-  # Callback which gets called after the ssr is saved to ensure that the
-  # past status is properly updated.  It should not normally be
-  # necessarily to call this method.
-  def update_past_status(identity)
-    old_status = self.past_statuses.last
-    if @prev_status and (not old_status or old_status.status != @prev_status)
-      self.past_statuses.create(status: @prev_status, date: Time.now, changed_by_id: identity.id)
+  # TODO: Verify that this method is no longer needed or being used
+  def candidate_statuses
```
wtholt: I just did a grep search and this method isn't called anywhere in the app. Probably safe to remove.

PR #748, opened 2016-11-17
KG - IE7 and Account Not Approved Modals

kyle-glick:
These modals appear in their respective situations and are not closable because the user should not be able to use the application.


kyle-glick:
https://www.pivotaltracker.com/story/show/128250443

config/locales/en.yml
```diff
@@ -402,6 +402,14 @@ en:
       header: "Add Primary PI"
       primary_pi: "Primary PI: *"
 
+  ########################
```
kyle-glick: I just noticed this should be removed. Please hold merge

PR #756, opened 2016-11-23
Default email in Contact Us form set to current user's and Feedback b…

jayhardee9:
…utton can hit a configurable URL.


app/controllers/contact_forms_controller.rb
```diff
@@ -19,9 +19,9 @@
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.~
 
 class ContactFormsController < ApplicationController
-
   def new
     @contact_form = ContactForm.new
+    @identity = current_identity
```
wtholt: No need for this, should have current_identity as a Devise method, right?

PR #769, opened 2016-11-30
wth - redirect to login, if identity is not signed in

wtholt:
When coming into proper for the first time, selecting a service, and
then clicking continue, the login page should be fired up instead of a
modal telling you that you are not signed in. [#135218283]


app/assets/javascripts/catalog.js.coffee
```diff
@@ -97,9 +97,8 @@ $(document).ready ->
   $(document).on 'click', '.submit-request-button', ->
     signed_in = parseInt($('#signed_in').val())
     if signed_in == 0
-      $('#modal_place').html($('#login-required-modal').html())
-      $('#modal_place').modal('show')
-      $('.modal #login-required-modal').removeClass('hidden')
+      window.location.replace('/identities/sign_in')
+      $('#flashes_container').html("<div class='alert alert-danger' role='alert'>You must be logged in</div>")
```
kyle-glick: Should YMLize the text here

PR #770, opened 2016-12-01
Fixing bug where line items were being locked in admin.

jleonardw9:


app/views/service_calendars/master_calendar/pppv/_template_line_items.html.haml
```diff
@@ -21,7 +21,7 @@
 - Dashboard::ServiceCalendars.pppv_line_items_visits_to_display(arm, service_request, sub_service_request, merged: merged, portal: portal, review: review, consolidated: consolidated).each do |org_name, livs|
   - line_items_visits << livs
   - ssr    = livs[0].sub_service_request
-  - locked = !ssr.can_be_edited? && !admin
```
kyle-glick: The admin variable is being used in the service calendar controller. Can you use the same logic in place of the session variable? If not, we will need to update all of the places where admin is used in the service calendar controller and service calendar views

PR #776, opened 2016-12-01
Jl arm name validation changes

jleonardw9:


app/assets/javascripts/service_details.js.coffee
```diff
@@ -49,3 +49,6 @@ $(document).ready ->
 
   $('#arms-table').on 'all.bs.table', ->
     $('.screening-info').tooltip()
+
+  $('#arms-table').on 'all.bs.table', ->
+    $('.name-validation').tooltip()
```
kyle-glick: Can we just add this after line 51 within the first handler block?

app/models/arm.rb
```diff
@@ -44,7 +44,8 @@ class Arm < ActiveRecord::Base
   after_save :update_liv_subject_counts
 
   validates :name, presence: true
-  validates_format_of :name, with: /\A([A-Za-z0-9][A-Za-z0-9]*([ ][A-Za-z0-9])?)*\z/
+  validates :name, format: { with: /\A([A-Za-z0-9][A-Za-z0-9]*([ ][A-Za-z0-9])?)*\z/, 
+                   message: "can not contain any of the following characters: [ ] * / \\ ? : " }
```
kyle-glick: I think you can YMLize the error message. See ActiveRecord I18n documentation
http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models

PR #780, opened 2016-12-01
KG - Service Search No Results

kyle-glick:
The controller code was causing the No Results to be displayed like a service. Typeahead has the custom template notFound that we can use instead. We can render a quick Handlebar template for that case.

### DIFF
--- a/app/assets/javascripts/catalog.js.coffee
+++ b/app/assets/javascripts/catalog.js.coffee
@@ -81,11 +81,13 @@ $(document).ready ->
                                           <span><strong>Abbreviation: {{abbreviation}}</strong></span><br>
                                           <span><strong>CPT Code: {{cpt_code}}</strong></span>
                                         </button>')
+        notFound: Handlebars.compile('<div class="tt-suggestion">No Results</div>')
       }
     }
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/catalog.js.coffee
```diff
@@ -81,11 +81,13 @@ $(document).ready ->
                                           <span><strong>Abbreviation: {{abbreviation}}</strong></span><br>
                                           <span><strong>CPT Code: {{cpt_code}}</strong></span>
                                         </button>')
+        notFound: Handlebars.compile('<div class="tt-suggestion">No Results</div>')
```
jayhardee9: Do you have to use Handlebars here?

app/assets/javascripts/catalog.js.coffee
```diff
@@ -81,11 +81,13 @@ $(document).ready ->
...
       }
     }
   ).on('typeahead:render', (event, a, b, c) ->
     $('[data-toggle="tooltip"]').tooltip()
   ).on('typeahead:select', (event, suggestion) ->
+    asd
```
jayhardee9: Is this a typo?

PR #783, opened 2016-12-05
Sj service request export changes

Stuart-Johnson:


app/helpers/service_calendar_helper.rb
```diff
@@ -136,6 +136,12 @@ def display_total_cost_per_study_otfs service_request, line_items
   #############################################
   # Grand Totals
   #############################################
+  def display_ssr_grand_total sub_service_request
+    sum = 0
```
jayhardee9: Is this assignment necessary?

app/views/service_requests/show.xlsx.axlsx
```diff
@@ -174,6 +174,9 @@ wb.add_worksheet(name: "Review") do |sheet|
           line_items_visit.line_item.pricing_scheme = 'effective' if @admin_offset
 
           service_row << display_visit_based_direct_cost(line_items_visit)
+          puts '#' * 50
```
jayhardee9: `puts`

app/helpers/service_calendar_helper.rb
Stuart-Johnson: I'm not 100% sure, but if you look at the file, it's written the same way all the other methods there are, I was going for consistency (and they all need revisited anyways, as another story).

PR #799, opened 2016-12-08
Jjh session

jayhardee9:
This pull request removes `sub_service_request_id` and `service_request_id` from the session variable. 
I placed a disabled hidden field named `service_request_id` in the application template for JS event handlers to pull from when making AJAX requests, if needed, for convenience. The base `ApplicationController` still sets `@service_request` and `@sub_service_request` as before. 
From now on, we need to be sure to pass along `(sub_)service_request_id`, if the receiving controller needs it.
If you see errors like: ```ActiveRecord::RecordNotFound (Couldn't find ServiceRequest with 'id'=)```, then you should just tack on `service_request_id=whatever` to your AJAX request, and that should do it.


jayhardee9:
Found some bugs in the service calendar under this branch, so I'm closing this PR while I fix them.

PR #802, opened 2016-12-08
Fixed parent orgs not updating children orgs [#135366883]

BigChiefSmidgeums:


app/models/organization.rb
```diff
@@ -107,6 +107,7 @@ def process_ssrs_parent
   end
 
   def update_ssr_org_name
+    self.org_children.each{ |org| org.update_ssr_org_name }
```
kyle-glick: Don't we also need to grab any parent organizations potentially and make sure to do the same? 

I think we can also do this a little more efficiently. I think this would do the same thing as both lines.

SubServiceRequest.where(organization: self.parents + all_child_organizations).each(:&update_org_tree)

jayhardee9: `all_child_organizations` will fetch every descendant of `self`, while `org_children` only returns immediate children. @BigChiefSmidgeums which should it be?

PR #817, opened 2016-12-13
KG - Unit Type # / Qty Type # Continuous Spinning Bug

kyle-glick:


app/controllers/line_items_controller.rb
```diff
@@ -32,7 +32,7 @@ def update
     if @line_item.update_attributes(params[:line_item])
       @service_request.update_attributes(status: 'draft')
       @line_item.sub_service_request.update_attributes(status: 'draft')
-      render partial: 'service_calendars/update_service_calendar'
+      render nothing: true
```
kyle-glick: Because the call is made via JSON, the javascript gets rendered as plain text. This is how I did the rest of the fields.

PR #830, opened 2016-12-20
Fixed the questionnaire controller's create action [#136213257]

BigChiefSmidgeums:


app/controllers/additional_details/questionnaires_controller.rb
```diff
@@ -26,7 +26,7 @@ def create
   end
 
   def update
-    @questionnaire.update_attributes(questionnaire_params)
+    @questionnaire.attributes = questionnaire_params
     if @questionnaire.save
```
jayhardee9: I think what you're wanting to do is `if @questionnaire.update(questionnaire_params)`. 
`attributes =` will call `assign_attributes` without any sanitation of the parameters, so it's possible to bypass mass-assignment security.

app/controllers/additional_details/questionnaires_controller.rb
```diff
@@ -26,8 +26,7 @@ def create
   end
 
   def update
-    @questionnaire.update_attributes(questionnaire_params)
-    if @questionnaire.save
+    if @questionnaire.update(questionnaire_params)
```
kyle-glick: Because .update returns the object, which isn't needed here, it might be better to just use .update_attributes.

http://stackoverflow.com/questions/27684038/rails-paperclip-update-vs-update-attributes

BigChiefSmidgeums: .update_attributes is an alias for .update as of 4.0.2
http://apidock.com/rails/v4.0.2/ActiveRecord/Persistence/update_attributes

PR #844, opened 2016-12-29
In the protocol filter in Dashboard, sort organizations by name [#136781729]

jayhardee9:


lib/dashboard/grouped_organizations.rb
```diff
@@ -30,6 +30,9 @@ def collect_grouped_options
       groups = @organizations.group_by(&:type)
       options = ["Institution", "Provider", "Program", "Core"].map do |type|
         next unless groups[type].present?
+
+        # Sort organizations by name
+        groups[type].sort! { |lhs, rhs| lhs.name <=> rhs.name }
```
kyle-glick: If @organizations is an ActiveRecord::Relation, could we sort in the group_by instead of here?

lib/dashboard/grouped_organizations.rb
```diff
@@ -30,6 +30,9 @@ def collect_grouped_options
       groups = @organizations.group_by(&:type)
```
kyle-glick: groups = @organizations.order(:name).group_by(&:type)

lib/dashboard/grouped_organizations.rb
jayhardee9: I wish @organizations was an AR relation, but the ProtocolsController passes in an Array of Organizations rather than an AR relation. But we can still sort before the group_by! Better to sort once than up to four times.

PR #851, opened 2017-01-03
Needed to require email to prevent devise registration error.

jleonardw9:


config/locales/devise.en.yml
```diff
@@ -75,7 +75,7 @@ en:
       last_name: "Last Name:*"
       first_name: "First Name:*"
       company: "Company:"
-      email: "Email:"
+      email: "Email:*"
```
kyle-glick: This might be out of scope, but we can use the class "required" on labels to add "*"s to fields instead of placing them in the YML. We do it in several other places currently. You can call the javascript method "set_required_fields()" to do this.

jleonardw9: Sounds good. I'm going to leave this PR how it is for now, but I'm going to put a story in to do this throughout the app.

PR #855, opened 2017-01-03
Other methods in ServiceCalendarsController needed @service_request initialized. [#136963065]

jayhardee9:
Needs to be cherrypicked:
3769f2a16057b14a1c3eca92cc40d04c03c2eca8
cc527cf0fea44450aa52a945ce6f9ca84bff9c37
3769f2a16057b14a1c3eca92cc40d04c03c2eca8
9e23f649bcbbafa4810bf1436bf6d470dd44ac0e


app/controllers/service_calendars_controller.rb
```diff
@@ -73,6 +73,7 @@ def update
   end
 
   def table
+    initialize_service_request
```
kyle-glick: The reason I added the procs to the before filters was that this was causing an issue before with authorization errors in the Dashboard. Instead, you can assign @service_request in the before_filter "authorize_dashboard_access", or more specifically I think, "authorize_protocol"

PR #856, opened 2017-01-03
KG - "Locked" Visual for Completed SSRs

kyle-glick:


app/helpers/service_requests_helper.rb
```diff
@@ -57,7 +57,7 @@ def organization_description_display(organization)
 
   def ssr_name_display(sub_service_request)
     header  = content_tag(:span, sub_service_request.organization.name + (sub_service_request.ssr_id ? " (#{sub_service_request.ssr_id})" : ""))
-    header += content_tag(:span, '', class: 'glyphicon glyphicon-lock locked') if !sub_service_request.can_be_edited?
+    header += content_tag(:span, '', class: 'glyphicon glyphicon-lock locked') if !sub_service_request.can_be_edited? || sub_service_request.is_complete?
```
jayhardee9: `!sub_service_request.can_be_edited? || sub_service_request.is_complete?` should be pulled into a helper method or a method on `SubServiceRequest` called, say, `locked?`. That way, it'll be easier to change the concept of being "locked" by not having to touch seven different files.

kyle-glick: @jayhardee9 
Can you take a look at the new changes? Because this logic is related to the editable statuses, I was able to move the logic into the can_be_edited? method.

app/models/sub_service_request.rb
```diff
@@ -320,7 +320,7 @@ def is_locked?
   def can_be_edited?
     if organization.has_editable_statuses?
       self_or_parent_id = find_editable_id(self.organization.id)
-      EDITABLE_STATUSES[self_or_parent_id].include?(self.status)
+      EDITABLE_STATUSES[self_or_parent_id].include?(self.status) && !is_complete?
```
jayhardee9: `complete` is an editable status for a couple Organizations, so this shouldn't be here.

kyle-glick: @jayhardee9 
According to Wenjun, even if it is an editable status, the users should not be able to edit a Complete SSR, no matter what. She said that instead, it means that users should be able to draft a new SSR for that organization.

jayhardee9: 10-4

PR #858, opened 2017-01-04
wth - RMID Functionalities - Part 1

wtholt:
Add Research Master ID field and validate field to be number only and
for presence if the Human Subjects checkbox is checked. [#136737011]

### DIFF
--- a/app/assets/javascripts/protocol_form.js.coffee
+++ b/app/assets/javascripts/protocol_form.js.coffee
@@ -19,6 +19,15 @@
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 $(document).ready ->
+
+  $(document).on 'click', '.human-subjects', ->
+    if $('.rm-id').hasClass('required-field')
+      $('.rm-id').removeClass('required-field')
+      $('.has-human-subject-info').val('false')
+    else
+      $('.rm-id').addClass('required-field')
+      $('.has-human-subject-info').val('true')
+
   # Protocol Edit Begin
   $(document).on 'click', '#protocol-type-button', ->
     protocol_id = $(this).data('protocol-id')
--- a/app/assets/stylesheets/protocol_form.sass
+++ b/app/assets/stylesheets/protocol_form.sass
@@ -46,3 +46,7 @@ textarea
 
 .gray_note 
   color: gray
+
+label.required-field:after
+  content: "*"
+
--- a/app/models/protocol.rb
+++ b/app/models/protocol.rb
@@ -100,6 +100,12 @@ class Protocol < ActiveRecord::Base
   attr_accessible :type
   attr_accessible :udak_project_number
   attr_accessible :vertebrate_animals_info_attributes
+  attr_accessible :research_master_id
+  attr_accessible :has_human_subject_info
+
+  validates :research_master_id, numericality: { only_integer: true }, allow_blank: true
+
+  validates :research_master_id, presence: true, if: :has_human_subject_info?
 
   attr_accessor :requester_id
   attr_accessor :validate_nct
@@ -128,6 +134,10 @@ class Protocol < ActiveRecord::Base
   validation_group :user_details do
     validate :validate_proxy_rights
     validate :primary_pi_exists
+  end
+
+  def has_human_subject_info?
+    self.has_human_subject_info == true
   end
 
   scope :for_identity, -> (identity) {
--- a/app/views/protocols/form/study_form_sections/_research_involving.html.haml
+++ b/app/views/protocols/form/study_form_sections/_research_involving.html.haml
@@ -29,7 +29,7 @@
       .form-group.row.human_subjects
         = rt_form.label :human_subjects, t(:protocols)[:studies][:research_involving][:humans][:header], class: 'col-lg-2 control-label'
         .col-lg-1
-          = rt_form.check_box :human_subjects, class: 'form-control'
+          = rt_form.check_box :human_subjects, class: 'form-control human-subjects'
 
       / Human Subjects selected begin
       .human_subjects_dependent{ display_if(rt_form.object.human_subjects) }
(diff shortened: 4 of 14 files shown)


app/models/protocol.rb
```diff
@@ -260,6 +266,10 @@ def active?
     study_type_question_group.nil? ? false : study_type_question_group.active
   end
 
+  def has_human_subject_info?
```
jayhardee9: ActiveRecord provides `#human_subject_info?`, so I think this method is a little redundant.

app/assets/stylesheets/protocol_form.sass
```diff
@@ -46,3 +46,7 @@ textarea
 
 .gray_note 
   color: gray
+
+label.required-field:after
```
kyle-glick: I actually like what you did here. I wasn't aware of this functionality. There's already a class for required fields, ".required", but it didn't have this css being used on it, and instead used a function in global.js.coffee to add the *s. Could you change this class to .required and perhaps remove the javascript function and calls to it?

app/models/protocol.rb
```diff
@@ -100,6 +100,12 @@ class Protocol < ActiveRecord::Base
   attr_accessible :type
   attr_accessible :udak_project_number
   attr_accessible :vertebrate_animals_info_attributes
+  attr_accessible :research_master_id
+  attr_accessible :has_human_subject_info
+
+  validates :research_master_id, numericality: true, allow_blank: true
```
kyle-glick: Would this allow non-integers to be entered? Shouldn't we use numericality: { only_integer: true }

app/views/protocols/form/study_form_sections/_research_involving.html.haml
```diff
@@ -29,7 +29,7 @@
       .form-group.row.human_subjects
         = rt_form.label :human_subjects, t(:protocols)[:studies][:research_involving][:humans][:header], class: 'col-lg-2 control-label'
         .col-lg-1
-          = rt_form.check_box :human_subjects, class: 'form-control'
+          = rt_form.check_box :human_subjects, class: 'form-control human-subjects'
```
kyle-glick: Instead of adding the class, you can select using $("#protocol_research_types_info_attributes_human_subjects")

app/assets/javascripts/protocol_form.js.coffee
```diff
@@ -19,6 +19,15 @@
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 $(document).ready ->
+
+  $(document).on 'click', '.human-subjects', ->
```
kyle-glick: Select using $("#protocol_research_types_info_attributes_human_subjects")

app/assets/javascripts/protocol_form.js.coffee
```diff
@@ -19,6 +19,15 @@
...
+    if $('.rm-id').hasClass('required-field')
```
kyle-glick: See comments about ".required"

app/views/protocols/form/study_form_sections/_research_involving.html.haml
kyle-glick: Needs to be replicated in "app/views/dashboard/protocols/form/study_form_sections/_research_involving.html.haml"

app/views/protocols/form/study_form_sections/_study_information.html.haml
```diff
@@ -27,6 +27,13 @@
       = t(:protocols)[:studies][:information][:subtext]
   .panel-body.container-fluid
     .form-group.row
+      = form.label :research_master_id, class: 'col-lg-2 control-label rm-id' do
```
kyle-glick: Needs to be replicated in "app/views/dashboard/protocols/form/study_form_sections/_study_information.html.haml"

app/views/protocols/form/study_form_sections/_study_information.html.haml
```diff
@@ -27,6 +27,13 @@
...
+        = link_to t(:protocols)[:studies][:information][:research_master_id], RESEARCH_MASTER_LINK
+      .col-lg-10
+        = form.text_field :research_master_id, class: 'form-control'
+        = form.hidden_field :has_human_subject_info, value: false, class: 'has-human-subject-info'
```
kyle-glick: What is the purpose of this hidden field?

db/migrate/20170103155625_add_research_master_id_to_protocols.rb
```diff
@@ -0,0 +1,5 @@
+class AddResearchMasterIdToProtocols < ActiveRecord::Migration
+  def change
+    add_column :protocols, :research_master_id, :integer
```
kyle-glick: Should there also be an index added here?

db/migrate/20170103201850_add_has_hs_info_to_protocols.rb
```diff
@@ -0,0 +1,5 @@
+class AddHasHsInfoToProtocols < ActiveRecord::Migration
+  def change
+    add_column :protocols, :has_human_subject_info, :boolean
```
kyle-glick: This is unnecessary coupling and could lead to data inconsistencies. We already have access to either protocol.research_types_info.human_subjects? or protocol.human_subjects.present?

app/views/protocols/form/study_form_sections/_research_involving.html.haml
wtholt: I think adding a class is cleaner. No real difference here.

app/views/protocols/form/study_form_sections/_study_information.html.haml
wtholt: Have to add validation to a RMID on the fly. This is a hidden field set to default to false. If a user checks the Human Subjects Info checkbox, sets this field to true which then will require presence validation on the RMID attribute.

db/migrate/20170103155625_add_research_master_id_to_protocols.rb
wtholt: No association

PR #866, opened 2017-01-06
SAW - Arm Name "2 and" (regex validation) Bug

SamWord:
Changed the regex to validate name format so that it accepts any character C where C is a Uppercase letter, lowercase letter, digit, or ~ ` ! @ # $ % ^ & ( ) _ - + = { } | < > , . ; " '
and the string is in the format C C C ... C

Moved the name format validation to a custom validation so that "Arm Name can not contain any of the following characters: [ ] * / \ ? :" error is not added when the arm name is blank.

Added specs for Arm validations


app/models/arm.rb
```diff
@@ -44,13 +44,18 @@ class Arm < ActiveRecord::Base
...
-  validates :name, format: { with: /\A([A-Za-z0-9][A-Za-z0-9]*([ ][A-Za-z0-9])?)*\z/, 
-                   message: "can not contain any of the following characters: [ ] * / \\ ? : " }
+  validate :name_formatted_properly
   validate :name_unique_to_protocol
 
   validates :visit_count, numericality: { greater_than: 0 }
   validates :subject_count, numericality: { greater_than: 0 }
 
+  def name_formatted_properly
+    if !name.blank? && name.match(/\A([A-Za-z0-9`~!@#$%^&()\-_+={}|<>.,;'"]+(([ ][A-Za-z0-9`~!@#$%^&()\-_+={}|<>.,;'"]*)*)?)\z/).nil?
```
jayhardee9: I think this regex is equivalent, while being simpler: `\A([A-Za-z0-9``~!@#$%^&()\-_+={}|<>.,;'"][ ]*)+\z`

SamWord:
I used Jay's regex, but added a `[ ]*` before the characters section so that the regex will only fail if it contains bad characters and added a `name=` method to address excess whitespace in a name.

PR #867, opened 2017-01-09
Kg - Additional Details Complete Form Rework

kyle-glick:


kyle-glick:
Before:

After:
screen shot 2017-01-05 at 9 39 44 am

app/controllers/additional_details/submissions_controller.rb
```diff
@@ -44,6 +44,8 @@ def create
     @submissions = @protocol.submissions
     @line_item = LineItem.find(submission_params[:line_item_id])
     @service_request = @line_item.service_request
+    @permission_to_edit = @protocol.project_roles.where(identity: current_user, project_rights: ['approve', 'request']).any?
```
jayhardee9: This logic is already in the Identity model as `has_correct_project_role?`, so `@permission_to_edit = current_user.has_correct_project_role?(@service_request)` would be better here.

app/models/protocol.rb
```diff
@@ -508,6 +508,20 @@ def has_non_first_draft_ssrs?
     sub_service_requests.where.not(status: 'first_draft').any?
   end
 
+  def has_incomplete_additional_details?
+    if service_requests.any? { |sr| sr.additional_detail_services.present? }
+      service_requests.each do |sr|
```
jayhardee9: I think it would be a good idea to get rid of this nested loop. Consider adding `has_many :line_items, through: :service_requests` to the Protocol model. Then you could just write something like `line_items.any?(&:has_incomplete_additional_details?)`.

app/views/dashboard/associated_users/destroy.js.coffee
```diff
@@ -32,7 +32,7 @@ $("#documents-table").bootstrapTable()
 $(".service-requests-table").bootstrapTable()
 
 $('.service-requests-table').on 'all.bs.table', ->
-	$(this).find('.selectpicker').selectpicker() #Find descendant selectpickers
+       $(this).find('.selectpicker').selectpicker() #Find descendant selectpickers
```
jayhardee9: Should this be two spaces?

spec/controllers/additional_details/submissions/delete_destroy_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+          protocol_id: @protocol.id
+        }
+      end
+
+      it 'should assign @protocol' do
+        expect(assigns(:protocol)).to eq(@protocol)
+      end
+
+      it 'should assign @submissions' do
+        expect(assigns(:submissions)).to be
```
jayhardee9: `eq(@submissions)`

app/controllers/additional_details/submissions_controller.rb
kyle-glick: @jayhardee9 
I originally planned on refactoring this code segment and other uses of it in the Dashboard, but that was a bit much for this PR. This is the way we've done it in the rest of the Dashboard code, so for now, I think keeping the uniformity might be better.

app/views/dashboard/associated_users/destroy.js.coffee
kyle-glick: @jayhardee9 
I have absolutely no clue. I couldn't get rid of this edit. I tried changing it to multiple space lengths and couldn't get it to be removed. I have a refactor coming right after this PR so I will make that update then.

app/controllers/additional_details/submissions_controller.rb
jayhardee9: Gotchya. Well let's change all those over to the Identity method in your upcoming refactor.

kyle-glick: @jayhardee9 
Agreed

PR #873, opened 2017-01-10
wth - RMID functionalities part 2

wtholt:
Added validation to make sure that RMID record exists and also added
validation to make sure that no other SPARC protocols are using the same
RMID record. [#136739621]


.env.example
```diff
@@ -0,0 +1,2 @@
+RM_ID_API='https://rmid.musc.edu/api/research_masters.json'
```
jayhardee9: Why not use one of the YAML files we already have in config?

PR #874, opened 2017-01-10
Jtm adding protocol filter specs

BigChiefSmidgeums:


spec/controllers/dashboard/protocol_filters/get_new_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      log_in_dashboard_identity(obj: @user)
+    end
+
+    it "creates the appropriate project filter object" do
+      xhr :get, :new, filterrific: @filterrific
+      expect( assigns(:protocol_filter).attributes ).to eq( build(:protocol_filter, identity_id: @user.id, search_query: "").attributes )
+    end
+
+    it "adds the filter to the user's protocol filter list" do
```
jayhardee9: It may be better to test the last element of `@user.protocol_filters` for the expected ProtocolFilter, rather than just the number of ProtocolFilters on `@user`

spec/controllers/dashboard/protocol_filters/post_create_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      before(:each) do
+        @user = build(:identity, protocol_filter_count: 3)
+
+        @protocol_filter = build(:protocol_filter, identity_id: @user.id)
+
+        log_in_dashboard_identity(obj: @user)
+      end
+
+      it "creates a protocol filter record" do
```
jayhardee9: Ditto

spec/controllers/dashboard/protocol_filters/post_create_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        expect{ xhr :post, :create, protocol_filter: @protocol_filter.attributes }.to change{ ProtocolFilter.count }.by(1)
+      end
+
+      it "adds the filter to the user's protocol filter list" do
```
jayhardee9: Ditto

PR #912, opened 2017-05-30
Musc i18n mailer

loganfreeman:
In order to make the test more robust, I changed the rspec to use i18n.


loganfreeman:
updated the pull request to deal with unit_factor

```
unit_factor || 0
```

will result in 0 if unit_factor is 0, then resulting in division by zero exception

loganfreeman:
```ruby
oe.service_providers.map(&:identity_id)
```


should use `&:identity_id` instead of `&:id`

kyle-glick:
@loganfreeman 
Can you merge master into this branch and push it back up? It should fix the spec timeout.

loganfreeman:
@kyle-glick 
I merged sparc-request/master into musc-i18n-mailer and pushed it back.

PR #964, opened 2017-07-06
wth - (SPARCRequest & SPARCDashboard) View Details With No "Research …

wtholt:
…Involving"

Added another logic check to render alert since it was getting through
first check, often because there could be a Research Type Info object
with no dependent objects.
[#148342841]

Story - https://www.pivotaltracker.com/story/show/148342841

### DIFF
--- a/app/views/protocols/view_details/_research_involving.html.haml
+++ b/app/views/protocols/view_details/_research_involving.html.haml
@@ -27,162 +27,162 @@
     %h5
       %strong
         = t(:protocols)[:studies][:research_involving][:header]
-        
-  - if research_types_info.present?
-    - if research_types_info.human_subjects || research_types_info.vertebrate_animals || research_types_info.investigational_products || research_types_info.ip_patents
-      / Human Subjects
-      - if research_types_info.human_subjects
-        .panel-body
-          %table.table.table-bordered
-            %thead.default-header
-              %th
-                %h5.col-lg-12
-                  %strong
-                    = t(:protocols)[:studies][:research_involving][:humans][:header]
-            %tbody
-              %tr
-                %td
-                  %label.col-lg-4
-                    = t(:protocols)[:studies][:research_involving][:humans][:nct_number]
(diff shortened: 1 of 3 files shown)


app/views/protocols/view_details/_research_involving.html.haml
```diff
@@ -27,7 +27,7 @@
     %h5
       %strong
         = t(:protocols)[:studies][:research_involving][:header]
-        
+
   - if research_types_info.present?
```
kyle-glick: We could merge this if statement with the one below it (`if research_types_info.human_subjects || research_types_info.vertebrate_animals... etc.`) and use only one `else` block with the no research involving alert to avoid duplicated code.

app/views/protocols/view_details/_research_involving.html.haml
```diff
@@ -27,7 +27,7 @@
...
     - if research_types_info.human_subjects || research_types_info.vertebrate_animals || research_types_info.investigational_products || research_types_info.ip_patents
```
kyle-glick: We could make this a model method instead to minimize view logic

app/views/protocols/view_details/_research_involving.html.haml
wtholt: I thought the same thing, but I was wondering why they had the check for the parent object there

wtholt: Will we always be creating a Research Types Info?

kyle-glick: @wtholt It looks like there will not always be a Research Types info based on my current data. I don't think I was very clear though. I meant we can combine the two into `if research_types_info.present? && (research_types_info.human_subjects || research_types_info.vertebrate_animals || ... )`. That way even if research_types_info is nil, it won't check against the children types

wtholt: Ah, gotcha. That makes sense

PR #966, opened 2017-07-06
KG - Step 2B: Calendar Query and Rendering Need Improvement Part 4 - Subject Count

kyle-glick:
Pivotal:
https://www.pivotaltracker.com/story/show/148344231

Changes:
1. Instead of rerendering the entire arm, rerender only the totals for the line items visit, and the arm totals. Because we have to use JSON for xeditable's error handling, we can render the necessary partials as strings and replace them with a success call in the AJAX request
2. Move authorization methods for calendar controllers to ApplicationController to avoid duplication
3. Move LineItemsVisitsController out of the Dashboard scope for consistency with other calendar controllers. It's not Dashboard-specific.
4. Globalize Javascript `getSRId()` and `getSSRId()` methods. I was unable to access the getSSRId() within the context of the subject count handler which is where I realized the duplication.


app/controllers/line_items_visits_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  private
+
+  def line_items_visit_params
+    params.require(:line_items_visit).permit(:subject_count)
+  end
+
+  private
+
+  def in_dashboard?
```
wtholt: You already have these methods in ApplicationController. They should be available since you are inheriting from that controller.

kyle-glick: @wtholt Good catch. I meant to remove these.

PR #996, opened 2017-07-12
Dotenv version sitename

loganfreeman:
[Can I use dotenv in production?](https://github.com/bkeepers/dotenv)

We think dotenv is convenient to add some changes to a site. We have situations where a site might have some temporary changes and those changes are not meant to be applied to all sites.

For example, add a site name to the main page and a version to the footer.


sparc-request:
Hmmm, that could work.  It’s something we can definitely put up for discussion.  For now, you can of course utilize it for your instance.

Andrew

PR #997, opened 2017-07-12
KG - Step 2B: Calendar Query and Rendering Need Improvement Part 5 - OTF Quantities

kyle-glick:
**Pivotal:
https://www.pivotaltracker.com/story/show/148346409**

Quantity and Units Per Quantity both use LineItemsController#update. I added JSON to return the necessary partials to render the corresponding totals.


kyle-glick:
I would recommend waiting until #966 is merged in so that there is less code to review.

PR #1002, opened 2017-07-14
JM - (Production Bug) An Errno::ENOENT occurred in service_requests#c…

jwiel86:
…atalog [#148509407]

https://www.pivotaltracker.com/story/show/148509407

### DIFF
--- a/app/controllers/service_requests_controller.rb
+++ b/app/controllers/service_requests_controller.rb
@@ -388,24 +388,27 @@ class ServiceRequestsController < ApplicationController
 
       @events = []
       begin
-        #to parse file and get events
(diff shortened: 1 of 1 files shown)


amcates:
Good fix but FYI, this only happens when I first deploy.  Once the cronjob runs it creates the file.  Another option would be to fire off the cronjob on deploy.  This solution suites our needs so no need to look at my option.

jwiel86:
That's what Stuart was telling me, but when I talked to Wenjun it looked like it was happening several times a month.

amcates:
Yes, when we would deploy bug fixes.  No worries, good fix.  Makes it even better.

PR #1003, opened 2017-07-14
SAW - Another PermissibleValue method

SamWord:
[#147708807](https://www.pivotaltracker.com/story/show/147708807)

This method gets a hash of PermissibleValue keys as the keys and values as the values (what you would expect)

### DIFF
--- a/app/models/permissible_value.rb
+++ b/app/models/permissible_value.rb
@@ -9,4 +9,9 @@ class PermissibleValue < ApplicationRecord
   def self.get_value_list(category)
     PermissibleValue.where(category: category).pluck(:value)
   end
+
+  # Get a hash of PermissibleValue keys as they keys and values as values
+  def self.get_hash(category)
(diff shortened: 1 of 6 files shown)


app/models/permissible_value.rb
```diff
@@ -10,8 +10,13 @@ def self.get_value_list(category)
     PermissibleValue.where(category: category).pluck(:value)
   end
 
+  # Get a hash of PermissibleValue keys as they keys and values as values
+  def self.get_hash(category)
```
amcates: What is the difference between this method and the one below?  They do the same thing

kyle-glick: `#get_hash` uses the standard format { key => value } but `#get_value_key_hash` uses a { value => key } format

amcates: yes, and I was confused because they looked so similar.

PR #1019, opened 2017-07-21
KG/JM - (Production Bug) A NoMethodError occurred in protocols#new [#…

jwiel86:
…148823111] https://www.pivotaltracker.com/story/show/148823111

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -188,7 +188,7 @@ class ApplicationController < ActionController::Base
         return true
       end
     # the service request is in first draft and has yet to be submitted (catalog page only)
-    elsif @service_request.status == 'first_draft'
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -188,7 +188,7 @@ def authorize_identity
         return true
       end
     # the service request is in first draft and has yet to be submitted (catalog page only)
-    elsif @service_request.status == 'first_draft'
+    elsif @service_request.status == 'first_draft' && controller_name == 'service_requests' && action_name == 'catalog'
```
kyle-glick: For anyone curious, the reason this didn't work is because a `@service_request.status` could be `first_draft` in the catalog (homepage), protocol page (step 1), or creating a new protocol in proper. The old code would allow a user to access both step 1 and the new protocol form without logging in. Step 1 wasn't causing any issues from what I can tell, but the new protocol form broke because it called `current_user`

PR #1023, opened 2017-07-24
fixed missing inactive labels for non-clinical service on Step 2B and 4 [#149441331]

ying-pbrc:


ying-pbrc:
I don't see it being linked automatically to 
https://www.pivotaltracker.com/n/projects/1918597/stories/149441331

PR #1024, opened 2017-07-25
JM/KG - (Production Bug): An Icalendar::Values::Date::FormatError occurr…

jwiel86:
…ed in service_requests#catalog [#148701413] https://www.pivotaltracker.com/story/show/148701413


app/controllers/service_requests_controller.rb
```diff
@@ -426,6 +426,7 @@ def setup_catalog_calendar
           active_alert.save
           ExceptionNotifier::Notifier.exception_notification(request.env, e).deliver unless request.remote_ip == '128.23.150.107' # this is an ignored IP address, MUSC security causes issues when they pressure test,  this should be extracted/configurable
         end
+      rescue ArgumentError
```
kyle-glick: `ArgumentError` should probably be handled by the previous rescue block so that we will receive notifications for it.
`rescue Exception, ArgumentError => e`

PR #1025, opened 2017-07-26
fixed 'deletion of all services' email bug [#149521941]

ying-pbrc:
Pivotal
https://www.pivotaltracker.com/n/projects/1918597/stories/149521941


Stuart-Johnson:
@ying-pbrc Looks like you've got a spec failure in spec/lib/notifier_logic/ssr_deletion_emails_spec.rb

I've pulled down your branch and tested locally, and it's failing there as well, so it's not a Travis issue. If you could fix the spec and push it up, we can get this approved and merged in. Thanks.

PR #1028, opened 2017-07-27
JM - (SPARCRequest) Deletion of All Services and Resubmit USERS Email Bug: Part 2

jwiel86:
#149574117
https://www.pivotaltracker.com/story/show/149574117

### DIFF
--- a/app/helpers/notifier_helper.rb
+++ b/app/helpers/notifier_helper.rb
@@ -58,12 +58,11 @@ module NotifierHelper
     (status == "submitted" || status == "request_amendment") && role == 'none' && !notes.empty?
   end
 
-  def determine_ssr(last_change, action_name)
+  def determine_ssr(last_change, action_name, deleted_ssrs)
     ssr_id = last_change.audited_changes['sub_service_request_id']
     if last_change.action == 'destroy'
       # This condition signifies a deleted SSR
       if SubServiceRequest.where(id: ssr_id).empty? && action_name == 'notify_user'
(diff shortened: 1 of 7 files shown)


app/helpers/notifier_helper.rb
```diff
@@ -58,12 +58,11 @@ def display_notes?(status, role, notes)
     (status == "submitted" || status == "request_amendment") && role == 'none' && !notes.empty?
   end
 
-  def determine_ssr(last_change, action_name)
+  def determine_ssr(last_change, action_name, deleted_ssrs)
     ssr_id = last_change.audited_changes['sub_service_request_id']
     if last_change.action == 'destroy'
       # This condition signifies a deleted SSR
       if SubServiceRequest.where(id: ssr_id).empty? && action_name == 'notify_user'
-        deleted_ssrs = @service_request.deleted_ssrs_since_previous_submission(true)
```
jwiel86: Since adding delayed_job to emails, this method call wasn't capturing the deleted ssrs.  Instead we capture before putting into the delayed_job queue.  See next comment

app/lib/notifier_logic.rb
```diff
@@ -157,16 +157,16 @@ def send_user_notifications(request_amendment: false)
     else
       approval = false
     end
-  
+    deleted_ssrs = @service_request.deleted_ssrs_since_previous_submission(true)
```
jwiel86: Here we capture the deleted_ssrs and pass them along.

PR #1029, opened 2017-07-27
KG/JM - (Production Bug) A NoMethodError occurred in protocols#new [#…

jwiel86:
…148823111] https://www.pivotaltracker.com/story/show/148823111


kyle-glick:
The problem was that we had originally changed the logic to only allow the user to load the service catalog, but hadn't considered any of the other actions that happen there.

Routes that should be doable:
`ServiceRequestsController#catalog`
`ServiceRequestsController#add_service`
`ServiceRequestsController#remove_service`
`CatalogsController#update_description`

Routes that should be blocked:
`ServiceRequestsController#protocol`
`ProtocolsController#new`

PR #1033, opened 2017-07-27
Jl - approved subsidy cost bug

jleonardw9:
When viewing an approved subsidy the calculation for pi_contribution was using the total cost, and not the cost at approval.
PR: https://www.pivotaltracker.com/n/projects/1918597/stories/149603327


jleonardw9:
Closing this for now. I should have based this off of master.

PR #1036, opened 2017-07-28
KG - (SPARCDashboard-Production Bug) View Consolidated Request Switch Visits

kyle-glick:
Pivotal:
https://www.pivotaltracker.com/story/show/149696355

**Note: Most of this code was pulled from v3.0.0 where this change has been made already, however I found that the Step 4 (Review) dropdown did not work and fixed it as well.**

### DIFF
--- a/app/controllers/service_requests_controller.rb
+++ b/app/controllers/service_requests_controller.rb
@@ -132,8 +132,8 @@ class ServiceRequestsController < ApplicationController
     @review       = true
     @portal       = false
     @admin        = false
-    @merged       = false
-    @consolidated = true
+    @merged       = true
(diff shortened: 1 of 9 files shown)


app/controllers/service_requests_controller.rb
```diff
@@ -132,8 +132,8 @@ def review
     @review       = true
     @portal       = false
     @admin        = false
-    @merged       = false
-    @consolidated = true
+    @merged       = true
+    @consolidated = false
```
kyle-glick: Bug: Review uses the merged calendar, i.e. the "Calendar Tab" calendar, not the consolidated calendar.

PR #1039, opened 2017-07-31
Jtm fixing status bug

BigChiefSmidgeums:


app/models/organization.rb
```diff
@@ -357,7 +359,7 @@ def setup_available_statuses
   def get_available_statuses
     tmp_available_statuses = self.available_statuses.reject{|status| status.new_record?}
     statuses = []
-    if tmp_available_statuses.empty?
+    if tmp_available_statuses.empty? || tmp_available_statuses.collect{|status| status.status} == DEFAULT_STATUSES.collect{|k, v| k}
```
kyle-glick: It may be shorter to do `tmp_available_statuses.map(&:status) == DEFAULT_STATUSES.keys`

wtholt: #collect and #map are the same thing (https://stackoverflow.com/questions/5254732/difference-between-map-and-collect-in-ruby). I also think DEFAULT_STATUSES is an array.

BigChiefSmidgeums: Thanks, Will! Works like a charm. I'll push the changes up.

PR #1061, opened 2017-08-15
wth - Milestones touchups

wtholt:
Please remove the "Study/ Project" wording from the Dates, to keep consistency with the corresponding labels on SPARCRequest Step 2 page. (sorry, my original mockup forgot the consistency with 2 of the labels...)
Please include a space between Service Requests table and Milestone table, currently they look right adjacent to each other;
Is there a way to make the date fields update real-time, instead of using the "Update Milestones" button? That could decrease the chance of losing data.

[#149694407]

Story - https://www.pivotaltracker.com/story/show/149694407

### DIFF
--- a/app/views/dashboard/protocols/show.html.haml
+++ b/app/views/dashboard/protocols/show.html.haml
@@ -25,7 +25,7 @@
   = render 'dashboard/service_requests/service_requests', protocol: @protocol, super_user_orgs: @super_user_orgs, permission_to_edit: @permission_to_edit, user: @user, view_only: false, show_view_ssr_back: @show_view_ssr_back, statuses_hidden: %(first_draft)
   .additional-details-submissions-panel
     = render "additional_details/submissions/submissions_panel", protocol: @protocol, submissions: @submissions
-  .milestones-panel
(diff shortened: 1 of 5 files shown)


app/views/dashboard/protocols/show.html.haml
```diff
@@ -25,7 +25,7 @@
   = render 'dashboard/service_requests/service_requests', protocol: @protocol, super_user_orgs: @super_user_orgs, permission_to_edit: @permission_to_edit, user: @user, view_only: false, show_view_ssr_back: @show_view_ssr_back, statuses_hidden: %(first_draft)
   .additional-details-submissions-panel
     = render "additional_details/submissions/submissions_panel", protocol: @protocol, submissions: @submissions
-  .milestones-panel
+  .milestones-panel.spacing-above
```
kyle-glick: Instead of adding a class to do this, moving the `panel` and `panel-default` div to this would do the same thing, using the built-in bootstrap `panel-group` styling.

wtholt: @kyle-glick I tried that, but it didn't add the spacing.

PR #1069, opened 2017-08-17
SJ - Orphaned service requests bug

Stuart-Johnson:
This makes service requests dependent destroy from protocols [#149729169]


kyle-glick:
Makes sense to me. I remember looking at that exact code a while back while talking to Wenjun and we found out that pretty much all relations to protocols and services requests are `dependent: destroy`, except for the relation between protocol and its service requests.

PR #1075, opened 2017-08-18
Jl- ssr status history recovery

jleonardw9:
Small task to recover the status and approval history of a deleted ssr and apply them to a new one.

https://www.pivotaltracker.com/n/projects/1918597/stories/150059945


kyle-glick:
This might be a good task to make generic and take input from so that it can be applied to more SubServiceRequests

PR #1077, opened 2017-08-18
wth - Refresh SPARC Protocol Information if associated to a validated…

wtholt:
… Research Master ID

Only display cue if RESEARCH_MASTER_ENABLED == true.

[#148336773]

Story - https://www.pivotaltracker.com/story/show/148336773

### DIFF
--- a/app/helpers/protocols_helper.rb
+++ b/app/helpers/protocols_helper.rb
@@ -51,12 +51,14 @@ module ProtocolsHelper
   end
 
   def display_rmid_validated_protocol(protocol, option)
-    if protocol.rmid_validated?
-      content_tag(
-        :h6,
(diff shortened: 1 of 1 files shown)


app/views/dashboard/protocols/_protocols_list.html.haml
```diff
@@ -73,7 +73,8 @@
             = protocol.id
         %td.title
           = short_title_display protocol
-          = display_rmid_validated_protocol(protocol, t('protocols.summary.rmid_short_title'))
+          - if RESEARCH_MASTER_ENABLED
```
kyle-glick: Since `if RESEARCH_MASTER_ENABLED` was added to `display_rmid_validated_protocol`, could we remove all of these condition statements in the views? It would remove a lot of duplication.

PR #1104, opened 2017-08-31
Lazy loading ldap user

loganfreeman:
1. add config entry `lazy_load` in application.yml ( LAZY_LOAD constant added in 01_obis_setup.rb )

2. changed directory search to use LAZY_LOAD.

```ruby
  def self.search(term)
    # Search ldap (if enabled) and the database
    if USE_LDAP && !SUPPRESS_LDAP_FOR_USER_SEARCH
      if LAZY_LOAD
        return self.search_and_merge_ldap_and_database_results(term)
      else
        return self.search_and_merge__and_update_ldap_and_database_results(term)
      end

    else # only search database once
      return search_database(term)
    end
  end
```
#### search_and_merge_ldap_and_database_results
just merge the results from database and ldap.
#### search_and_merge__and_update_ldap_and_database_results
not only  merge the results from database and ldap, but also create database record if ldap result not found in database. *can create a lot of database record on the fly and be slow*

3. add a method suggestion_value to Identity model

```ruby
  def suggestion_value
    if LAZY_LOAD && USE_LDAP
      ldap_uid
    else
      id
    end
  end
```

4. add a method find_or_create to Identity model

```ruby
  def self.find_or_create(id)
    if LAZY_LOAD && USE_LDAP
      return Directory.find_or_create(id)
    else
      return self.find(id)
    end
  end
```

5. change to use `Identity.find_or_create` in associated_user_controller `new` method


Stuart-Johnson:
https://www.pivotaltracker.com/story/show/151338873

PR #1107, opened 2017-09-03
get ldap filter for full name

loganfreeman:
using `Net::LDAP::Filter.contains` to search in ldap is slow.

But the ldap filter lacks the capability to search by first name and last name.


Stuart-Johnson:
https://www.pivotaltracker.com/story/show/151338923

PR #1112, opened 2017-09-06
Permissible Values Implemented

SamWord:
[#147708807](https://www.pivotaltracker.com/story/show/147708807) which is a part of [##3615063 Configurations Generalization Epic](https://www.pivotaltracker.com/epic/show/3615063)

This pull request removed constant calls from a number of constants (see the changes to application.yml and 01_obis_setup.rb) and replaced them with PermissibleValue model method calls.


kyle-glick:
A few larger spec changes that we made in this:
1. Database cleaner was changed to use the `truncation` strategy, which is recommended by Capybara to use as the default cleaning strategy. See [here](https://stackoverflow.com/questions/10904996/difference-between-truncation-transaction-and-deletion-database-strategies).
Capybara Documentation:
```
# Transactional fixtures do not work with Selenium tests, because Capybara
# uses a separate server thread, which the transactions would be hidden
# from. We hence use DatabaseCleaner to truncate our test database.
```
2. We decided to build permissible values as a `before :suite` because they are mostly static values and building large quantities of them inside of each test was very costly on time.

PR #1116, opened 2017-09-07
SJ - Redo Jtm add form func to org

Stuart-Johnson:
Reverts the original revert, effectively a "redo."

Reverts sparc-request/sparc-request#1110


app/helpers/dashboard/sub_service_requests_helper.rb
```diff
@@ -208,27 +208,36 @@ def display_owner(ssr)
...
-                          data: {
-                            service_id: li.service.id,
-                            protocol_id: protocol.id,
-                            line_item_id: li.id
-                          }
-                        )
+      submissions = {}
+      if has_incomplete_service
+        submissions[:Services] = []
+        ssr.line_items.includes(:service).map(&:service).each do |service|
```
kyle-glick: `Service.where(id: ssr.line_items.pluck(:service_id)).distinct.each` would be a bit more efficient. Instead of hitting the database `n services` times, it hits it once.

app/views/additional_details/submissions/form_partials/_likert_form_partial.html.haml
```diff
@@ -19,10 +19,10 @@
 -# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.~
 
 %br
-- item.item_options.each_with_index do |item_option, index|
-  .likert-group.no-padding{ class: "col-sm-#{(12 / item.item_options.length).ceil}"}
+- item.item_options.each do |item_option|
+  .likert-group.no-padding{ class: "col-sm-#{(12 / item.item_options.size).ceil}"}
     .col-sm-12.text-center.no-padding
-      = index+1
+      = item.item_options.index(item_option)+1
```
kyle-glick: This is probably a bit more inefficient than using `each_with_index` because it will need to loop through `item_options` for each option. Is there a specific reason to change it?

app/helpers/dashboard/sub_service_requests_helper.rb
Stuart-Johnson: @kyle-glick the use of .includes(:services) means that the services are eager loaded with the line items. Tested in console, this creates only two queries. One to load all the line items, and one to load all the services. There is no query per service.

PR #1119, opened 2017-09-07
SW/KG - Permissible Values Implementation

kyle-glick:
[#147708807](https://www.pivotaltracker.com/story/show/147708807) which is a part of [##3615063 Configurations Generalization Epic](https://www.pivotaltracker.com/epic/show/3615063)

This pull request removed constant calls from a number of constants (see the changes to application.yml and 01_obis_setup.rb) and replaced them with PermissibleValue model method calls.

A few larger spec changes that we made in this:
1. Database cleaner was changed to use the `truncation` strategy, which is recommended by Capybara to use as the default cleaning strategy. See [here](https://stackoverflow.com/questions/10904996/difference-between-truncation-transaction-and-deletion-database-strategies).
**Capybara Documentation:**
```
# Transactional fixtures do not work with Selenium tests, because Capybara
# uses a separate server thread, which the transactions would be hidden
# from. We hence use DatabaseCleaner to truncate our test database.
```
2. We decided to build permissible values as a `before :suite` because they are mostly static values and building large quantities of them inside of each test was very costly on time. To do this, we had DatabaseCleaner exclude the `PermissibleValues` table from its cleaning.


kyle-glick:
Note: #1139 will conflict with this branch. When merged into `v3.0.0b`, this branch will need to be updated in several places.

`app/models/organization.rb#get_available_statuses`,
`app/models/organization.rb#get_editable_statuses`,
`app/views/catalog_manager/shared/_available_statuses.html.haml`,
`app/views/catalog_manager/shared/_status_options_box.html.haml`
- change `AVAILABLE_STATUSES` to `PermissibleValue.get_hash('status')`
- change `DEFAULT_STATUSES` to `PermissibleValue.get_key_list('status', true)`

`app/views/catalog_manager/shared/_available_statuses.html.haml`,
`app/views/catalog_manager/shared/_status_options_box.html.haml`
`AvailableStatus::TYPES[status]` to `PermissibleValue.get_value('status', status)`

PR #1135, opened 2017-09-22
fixed sending survey bugs - #151328240 

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/151328240

### DIFF
--- a/app/controllers/dashboard/sub_service_requests_controller.rb
+++ b/app/controllers/dashboard/sub_service_requests_controller.rb
@@ -84,7 +84,7 @@ class Dashboard::SubServiceRequestsController < Dashboard::BaseController
 
   def update
     if @sub_service_request.update_attributes(sub_service_request_params)
-      @sub_service_request.distribute_surveys if @sub_service_request.is_complete?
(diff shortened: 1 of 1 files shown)


app/controllers/dashboard/sub_service_requests_controller.rb
```diff
@@ -84,7 +84,7 @@ def show
 
   def update
     if @sub_service_request.update_attributes(sub_service_request_params)
-      @sub_service_request.distribute_surveys if @sub_service_request.is_complete?
+      @sub_service_request.distribute_surveys if (@sub_service_request.status == 'complete' && sub_service_request_params[:status].present?)
```
kyle-glick: I think instead of checking that the status is only 'complete', this should probably check if the status is in FINISHED_STATUSES

`if (FINISHED_STATUSES.include?(@sub_service_request.status) && sub_service_request_params[:status].present?)`

ying-pbrc:
It was sending survey notification if the status is in finished_statuses, however, that was causing some issues, see details in https://www.pivotaltracker.com/n/projects/1918597/stories/151310907.
So Wenjun asked to change it to only send survey with ‘complete” status:
https://www.pivotaltracker.com/n/projects/1918597/stories/151328240

PR #1149, opened 2017-10-04
Settings Implementation

kyle-glick:
Story:
https://www.pivotaltracker.com/story/show/147540421

All values from `application.yml` have been moved to the database as `Setting`s. Settings contain a `key`, `value`, `data_type`, `friendly_name`, `description`, `group`, `version`, `parent_key`, and `parent_value`.

The `key` is considered the primary key of a Setting. You find it by using `#find_by_key(<String>)`. We used a custom `value` override to return the value of a Setting based on the Setting's data_type.

The `friendly_name` and `description` will be used for the eventual User Interface. These provide human-readable names and descriptions of each Setting for users to understand what the Setting does.

The `group` attribute can be used to define groups of Settings, i.e. considering them a family of Settings. This could allow us to put similar settings together in the eventual User Interface.

The `version` attribute can be used to define versions of Settings. As of right now we do not have a use for this, but it was a requested column.

The `parent_key` and `parent_value` are used to define parent Settings. This defines a relationship where child Settings are dependent on a parent Setting being turned on or off to have the option to manipulate them. This is mostly for the eventual User Interface.

Major Changes:
1. Constant calls are replaced by the following format: `Setting.find_by_key("key").value`. This returns the value of the Setting in the correct data_type for that Setting.
2. The `settings` table is excluded from the DatabaseCleaner to allow for one-time initialization when running the test suite. This has a few implications:
- Several initializers had to use `begin-rescue`s due to Settings not being initialized yet in the test suite.
- Settings must be built after the DatabaseCleaner cleans the database before the suite. Because of potential concurrency issues, this needs to be done in the same `before :suite` block as the DatabaseCleaner (see `database_cleaner.rb`)
- Settings should not be initialized in the tests. Rather, they should be stubbed similar to how we currently do it. See `3.` for details.
3. `stub_const("CONST", "VALUE")` is no longer suitable in the specs. Instead, use `stub_config("key", value)` outside of an example block to stub a constant for all tests in that block. Note: The value needs to be passed in as the correct type. If the Setting is a boolean, value should be `true/false`. If the Setting is JSON, the value must be a valid JSON hash or array. Etc.


kyle-glick:
Note: #1148 will conflict with this branch. When merged into `v3.0.0b`, this branch will need to be updated and `USE_EPIC` in `app/lib/epic_queue_manager.rb` will need to be replaced with `Setting.find_by_key("use_epic").value`

PR #1152, opened 2017-10-12
JM - (SPARCRequest) SubServiceRequests without a Service Requester [#…

jwiel86:
…151725253]

https://www.pivotaltracker.com/story/show/151725253

### DIFF
--- a/app/lib/add_service.rb
+++ b/app/lib/add_service.rb
@@ -65,7 +65,7 @@ class AddService
       end
     end
     sub_service_request = service_request.sub_service_requests.create(
-      organization_id: organization.id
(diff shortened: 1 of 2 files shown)


app/lib/add_service.rb
```diff
@@ -65,7 +65,7 @@ def find_or_create_sub_service_request(line_item, service_request)
       end
     end
     sub_service_request = service_request.sub_service_requests.create(
-      organization_id: organization.id
```
kyle-glick: Because Rails can automatically assert IDs from class associations, you can shorten this statement to
```service_request.sub_service_requests.create(organization: organization, service_requester: @current_user)```

PR #1157, opened 2017-10-16
Add a funding source undetermined

loganfreeman:
`funding status => funding source => rate type`
`undetermined => undetermined => unfunded_rate_type`

not every protocol has a funding source.

### DIFF
--- a/config/constants.yml
+++ b/config/constants.yml
@@ -511,6 +511,7 @@ funding_sources: { 'College Department': 'college',
                    'Student Funded Research': 'unfunded' }
 
 funding_statuses: { 'Pending Funding': 'pending_funding',
+                    'Undetermined': 'undetermined',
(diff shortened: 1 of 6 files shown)


config/constants.yml
```diff
@@ -511,6 +511,7 @@ funding_sources: { 'College Department': 'college',
                    'Student Funded Research': 'unfunded' }
 
 funding_statuses: { 'Pending Funding': 'pending_funding',
+                    'Undetermined': 'undetermined',
```
kyle-glick: This will cause conflicts with the new PermissibleValues branch. Ideally this branch should wait until #1119 is merged so that we can use the updated PermissibleValues formatting

loganfreeman: Agree.

loganfreeman:
I will close this pr.

PR #1174, opened 2017-10-26
JM - (Production Bug)A NoMethodError occurred in protocols#create [#1…

jwiel86:
…52288500]

If testing manually, please see the story.  There are steps to reproduce. Thanks!

### DIFF
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -322,8 +322,10 @@ class ProtocolsController < ApplicationController
   def fix_identity
     attrs               = protocol_params
     attrs[:project_roles_attributes].each do |index, project_role|
-      identity = Identity.find_or_create project_role[:identity_id]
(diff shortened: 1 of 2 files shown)


app/controllers/protocols_controller.rb
```diff
@@ -322,8 +322,10 @@ def convert_date_for_save(attrs, date_field)
   def fix_identity
     attrs               = protocol_params
     attrs[:project_roles_attributes].each do |index, project_role|
-      identity = Identity.find_or_create project_role[:identity_id]
-      project_role[:identity_id] = identity.id
+      if project_role[:identity_id].present?
```
kyle-glick: I noticed this exact code is also in `Dashboard::ProtocolsController.rb`. This change probably needs to happen there as well.

jwiel86: Done! Thanks!

PR #1177, opened 2017-10-30
customize time zone with dotenv

loganfreeman:
# Note on dotenv load order
dotenv is initialized in your Rails app during the `before_configuration` callback, which is fired when the Application constant is defined in config/application.rb with `class Application < Rails::Application`. If you need it to be initialized sooner, you can manually call `Dotenv::Railtie.load`.

# use dotenv to load time_zone
```
config.time_zone = ENV['time_zone']
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,3 +3,4 @@ SPARC_VERSION=v3.0.0
 institution=University
 institution_logo=/assets/institution_logo2017.png
 org_logo=/assets/org_logo2017.png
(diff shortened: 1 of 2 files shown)


Stuart-Johnson:
@loganfreeman This needs a pivotal story.

kyle-glick:
@Stuart-Johnson 
Pivotal: https://www.pivotaltracker.com/story/show/157564966

PR #1178, opened 2017-11-06
jtm - (SPARCRequest & SPARCDashboard) Adding 'selected' column to available and editable status tables

BigChiefSmidgeums:


app/models/organization.rb
```diff
@@ -402,10 +375,9 @@ def has_tag? tag
 
   private
 
-  def create_editable_statuses
-    EditableStatus.statuses.each do |status|
-      self.editable_statuses.create(status: status)
-    end
+  def create_statuses
+    EditableStatus.import EditableStatus.types.map{|status| EditableStatus.new(organization: self, status: status)}
+    AvailableStatus.import AvailableStatus.types.map{|status| AvailableStatus.new(organization: self, status: status)}
```
kyle-glick: This doesn't appear to be correctly creating statuses on a new organization. Another thing I noticed is that it doesn't handle the "selected" attribute on the default statuses.
image

spec/factories/organization.rb
```diff
@@ -99,6 +99,11 @@
       end
     end
 
+    after(:create) do |organization, evaluator|
```
kyle-glick: I'm thinking we should avoid setting all statuses to `selected` because there are many test cases that rely on certain statuses being/not being selected. If Organizations create statuses with default statuses selected, then we can just update the statuses to selected in specs that expect them to be selected.

app/models/organization.rb
kyle-glick: My thought for how to set the `selected` attribute is something like this
```EditableStatus.types.map{|status| EditableStatus.create(organization: self, status: status, selected: AvailableStatus.defaults.include?(status))}```

I'm not actually sure what import does

**Comment by kyle-glick** [comment]
Another detail I noticed from manual testing is that there is a lot of cache hits being made to editable and available statuses. I'm not 100% sure if they're related to the status view or the model

File: app/views/catalog_manager/shared/_available_statuses.html.haml [hunk 3]
```diff
@@ -17,27 +17,24 @@
 -# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
 -# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
 -# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
 #all_statuses{style: @organization.use_default_statuses ? "display: none;" : "display: block;"}
```
kyle-glick: I noticed during manual testing that this page makes a large number of cache hits. We may be able to use a combination of eager loading and Ruby enumerable selection to improve performance.

BigChiefSmidgeums: The large number of cache hits is due to how we retrieve statuses from permissible values. Stuart informed me that we're going to change how we initialize variables from permissible values, so the issue will be solved when that is complete.

app/models/organization.rb
BigChiefSmidgeums: This is the correct behavior for when use_default_statuses is set to true for an organization.

PR #1184, opened 2017-11-07
SJ - Fixing indirect cost validation [#152641470]

Stuart-Johnson:
https://www.pivotaltracker.com/story/show/152641470

### DIFF
--- a/spec/models/protocol/valid?_spec.rb
+++ b/spec/models/protocol/valid?_spec.rb
@@ -30,10 +30,6 @@ RSpec.describe Protocol, type: :model do
   build_study_type_questions()
   build_study_type_answers()
 
-  before :each do
-    Setting.find_by_key("use_indirect_cost").update_attribute(:value, true)
-  end
-
   describe 'should validate funding source for projects' do
     it 'should raise an exception if funding source is nil' do
       project = Project.create(attributes_for(:protocol))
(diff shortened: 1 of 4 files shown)


wtholt:
@Stuart-Johnson Looks good, but I think you can checkout those db/schema.rb changes.

spec/models/protocol/valid?_spec.rb
```diff
@@ -38,6 +38,8 @@
     end
   end
 
-  it { is_expected.to validate_numericality_of(:indirect_cost_rate).is_greater_than_or_equal_to(1) }
-  it { is_expected.to validate_numericality_of(:indirect_cost_rate).is_less_than_or_equal_to(1000) }
+  if Setting.find_by_key('use_indirect_cost').value
```
kyle-glick: After discussing with @Stuart-Johnson, we should not use an `if` condition here, but instead a `before :each do` block where `use_indirect_cost` is turned on.

```
before :each do
  stub_config("use_indirect_cost", true)
end
```

kyle-glick:
Like @wtholt said and after discussing with @Stuart-Johnson, it seems like the schema changes should be checked out, though we may want to consider the effect of the inconsistent collation in the schema file. Is it a big deal or does it need to be updated? Schema does update itself automatically,but it only updates the information for specific tables being changed in migrations, so because a lot of older tables have not be changed in a while, the collation may be out-of-date.

PR #1186, opened 2017-11-08
update settings

loganfreeman:
add a script to update settings from json file.

the json file can be generated from rails console:

```
File.open('tmp/settings.json', 'w') { |file| file.write(Setting.all.to_json) }
```

then we can modify the json file and update the setting by:

```
script/update_setting tmp/settings.json
```


Stuart-Johnson:
@kyle-glick @loganfreeman Yeah, this should be a rake task. Also, this should link to a pivotal story.

PR #1187, opened 2017-11-08
if research master enabled

loganfreeman:
1. add a safe_value method to Setting class

```
Setting.safe_value('xxx') return nil

Setting.safe_value('use_cas') return false
```

2. check if research_master_enabled is enabled

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -32,8 +32,14 @@ class Setting < ApplicationRecord
 
   validate :value_matches_type, if: Proc.new{ !self.read_attribute(:value).nil? }
   validate :parent_value_matches_parent_type, if: Proc.new{ self.parent_key.present? }
-  
-  # Needed to correctly write boolean true and false as value in specs  
+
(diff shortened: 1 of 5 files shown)


app/models/setting.rb
```diff
@@ -32,8 +32,14 @@ class Setting < ApplicationRecord
 
   validate :value_matches_type, if: Proc.new{ !self.read_attribute(:value).nil? }
   validate :parent_value_matches_parent_type, if: Proc.new{ self.parent_key.present? }
-  
-  # Needed to correctly write boolean true and false as value in specs  
+
+  def self.safe_value(key)
```
kyle-glick: I like the idea of using this instead of `Setting.find_by_key("key").value` and running the risk of a nilvalue. It might be a good idea to make this our standard practice for retrieving settings. We could just call it `self.get_value(key)`

Stuart-Johnson:
@loganfreeman This needs a pivotal link.

PR #1190, opened 2017-11-20
using sanitized_data gem

amcates:


Stuart-Johnson:
This PR should fix this story: https://www.pivotaltracker.com/story/show/152339799

PR #1211, opened 2017-12-18
wth - (SPARCRequest & SPARCDashboard) Permissible Values should Follo…

wtholt:
…w the Sort_Order Defined

order by sort order in all places where we grab permissible values

[#152658805]

Story - https://www.pivotaltracker.com/story/show/152658805


wtholt:
@kyle-glick 

I agree, but we have some Permissible Values that do not have a sort_order. 

screen shot 2017-12-27 at 11 07 22 am

PR #1232, opened 2018-01-30
Jtm add quantity validation non clinical services

BigChiefSmidgeums:
Current, SPARCDashboard Admin Edit section is allowing admin users to add/edit Non-clinical services with blank "Quantity Requested" field, which is creating bad data and could cause silent failure when pushing the request to SPARCFulfillment.
Please:
1). Add validation to the "Quantity Requested" field for Adding/Editing Non-clinical services;
2). Please indicate the required fields with "*" in the popup modal.

### DIFF
--- a/config/locales/dashboard.en.yml
+++ b/config/locales/dashboard.en.yml
@@ -322,8 +322,8 @@ en:
         complete: "Complete"
         actions: "Actions"
       form:
-        service: "Service:"
-        quantity_requested: "Quantity Requested:"
+        service: "Service:*"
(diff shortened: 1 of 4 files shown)


config/locales/dashboard.en.yml
```diff
@@ -322,8 +322,8 @@ en:
         complete: "Complete"
         actions: "Actions"
       form:
-        service: "Service:"
```
kyle-glick: Rather than adding asterisks to the YML, this should be done by adding the `required` class to the form labels

PR #1265, opened 2018-02-28
extracting ldap.yml to settings table [#147540955]

amcates:
https://www.pivotaltracker.com/story/show/147540955

### DIFF
--- a/app/lib/directory.rb
+++ b/app/lib/directory.rb
@@ -21,32 +21,38 @@
 require 'net/ldap'
 
 class Directory
+  attr_accessor :ldap_host, :ldap_port, :ldap_base, :ldap_encryption, :domain, :ldap_uid, :ldap_last_name, :ldap_first_name, :ldap_email, :ldap_auth_username, :ldap_auth_password, :ldap_filter
 
-  begin
-    use_ldap = Setting.find_by_key("use_ldap").value || Rails.env == 'test'
-  rescue
-    use_ldap = true
-  end
+  def initialize
+    @ldap_host           = nil
+    @ldap_port           = nil
+    @ldap_base           = nil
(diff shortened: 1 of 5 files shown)


config/defaults.json
```diff
@@ -638,5 +638,125 @@
     "version":        "",
     "parent_key":     "",
     "parent_value":   ""
+  },
+  {
+    "key":            "ldap_host",
```
kyle-glick: For ease-of-use, can you move these so that they are alphabetical order? I also noticed the new funding settings aren't in alphabetical order, but we can fix that elsewhere.

app/lib/directory.rb
```diff
@@ -102,18 +108,18 @@ def self.find_or_create(ldap_uid)
   # Net::LDAP::Entry.
   def self.search_ldap(term)
```
kyle-glick: Because Directory was changed to be used as an initialized object all of the class methods in this file probably need to be changed to instance methods.

db/migrate/20180228194754_add_ldap_config_to_settings.rb
```diff
@@ -0,0 +1,53 @@
+class AddLdapConfigToSettings < ActiveRecord::Migration[5.1]
+  def up
```
kyle-glick: There's a file `app/lib/default_settings_populator` with the logic to create existing settings. Ideally we should open ldap config and populate the settings there in case someone needs to re-generate these settings. Because the setting keys are unique, it can be rerun and won't overwrite existing settings but will add new settings if any are found.

config/defaults.json
amcates: They can be moved but it won't matter in our database unless you have a fresh install.

PR #1283, opened 2018-04-02
KG - Display Order Not Updating Bug

kyle-glick:
https://www.pivotaltracker.com/story/show/155976063

This is covering Kyle Hutson's last comment about the Display Order not updating.

There were 2 issues present:
1. The JavaScript that is supposed to trigger when changing the select was not doing so. Something in the v3.2.0 update caused this to no longer trigger. I added the selector to the JavaScript and it now sends requests as it should.
2. The SurveyUpdaterController's params were not allowing `display_order` to be passed. I had to add this as well.


kyle-glick:
Closing this story to instead remove display order altogether.
See https://www.pivotaltracker.com/story/show/155976063/comments/187913763

PR #1299, opened 2018-04-13
JM - (SPARCForms) View Response Likert Bug

jwiel86:
https://www.pivotaltracker.com/story/show/156545721

When viewing a user's Survey or Form response with a Likert scale, the user is able to click other Likert options' labels and change the selected radio button. This will not change the data but the radio buttons should not be selectable.

### DIFF
--- a/app/views/surveyor/responses/form/_response_question.html.haml
+++ b/app/views/surveyor/responses/form/_response_question.html.haml
@@ -36,7 +36,7 @@
     .form-group
       %p.question-description
         = raw(question.description)
-  .question-options{ id: ['new', 'edit', 'preview'].include?(action_name) ? '' : 'disable-likert' }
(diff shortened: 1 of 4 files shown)


app/views/surveyor/responses/form/_response_question.html.haml
```diff
@@ -36,7 +36,7 @@
     .form-group
       %p.question-description
         = raw(question.description)
-  .question-options
+  .question-options{ id: ['new', 'edit', 'preview'].include?(action_name) ? '' : 'disable-likert' }
```
kyle-glick: This will give us multiple elements with the same ID which breaks DOM conventions. Can we use a class instead?

Stuart-Johnson: Agreed, this can cause lots of problems, especially in forms.

PR #1301, opened 2018-04-17
Ability to specify multiple ldap_base for searching [#152927779]

amcates:
This converts the ldap_base setting to a json array and uses that array to search multiple base DN.   This will still work with single base DN instances.


morizono:
This request is very happy making news--was one of the issues we had in our
environment when we started with SPARC way back when-- too many
heterogeneous Active Directories and the gem was unable to deal with them.

sparc-request:
This won’t solve the problem of multiple AD servers. Just multiple OU on the same server.  It allows you to search different branches of the tree.  I’m sure this could easily be refactored to allow multiple AD servers using a similar method the ldap_host setting.

PR #1311, opened 2018-05-03
KG - Remove Surveys Display Order

kyle-glick:
Pivotal: https://www.pivotaltracker.com/story/show/155976063 (See Comments)

In the comments of this story we decided that Display Order was not needed and sorting by title/version is instead a better way to go.


Stuart-Johnson:
@kyle-glick Please resolve the conflicts, thanks.

PR #1312, opened 2018-05-03
KG - Update version when changing access code

kyle-glick:
Pivotal: https://www.pivotaltracker.com/story/show/155990205

Changing a survey's access_code can cause 2 surveys to have the same version. This PR fixes this by updating the version to the n+1th version for that access_code on update.

### Before updating the access code
image

### After updating the access code
image


Stuart-Johnson:
@kyle-glick This one also has some conflicts.

PR #1314, opened 2018-05-04
KG - Permissible Values is_available

kyle-glick:
Pivotal: https://www.pivotaltracker.com/story/show/156618166

Add a field to PermissibleValues that allows us to filter out "old" values that aren't needed any more. When loading PermissibleValues via the various class methods, it will only pull available values.


kyle-glick:
**Help Wanted:** Migrations are failing because an OLD migration is referencing the changed code which is expecting `permissible_values.is_available` to be a table attribute. What is the best way to go about fixing this? Should I edit the old migration to manually get the status hash, rather than going through the PermissibleValue methods?

PR #1320, opened 2018-05-14
KG - Refactor Settings Retrieval

kyle-glick:
**Pivotal: https://www.pivotaltracker.com/story/show/157564028**

**Original PR: https://github.com/sparc-request/sparc-request/pull/1187**

Created class method `get_value(key)` on `Setting` that provides a friendlier way to get the value of a Setting. I then replaced all instances of `Setting.find_by_key(some_string).value` and `Setting.find_by_key(some_string).try(:value)`.


amcates:
From what I can tell.  The only real change here is that get_value will return nil if the setting has no value or does not exist

PR #1336, opened 2018-05-29
Funding documents table improvement [#157677166]

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/157677166

merge PR#1335 first


ying-pbrc:
bootstrap-table-export is an extension of TableExport jquery plugin. bootstrap-table-export won't work if missing any required dependency.
See https://github.com/wenzhixin/bootstrap-table-examples/blob/master/extensions/export.html#L13

PR #1356, opened 2018-06-11
this fixes a deprecation warning when it tries to load the accordian …

amcates:
…list [#158261574]


amcates:
This is actually included in https://github.com/sparc-request/sparc-request/pull/1357

PR #1399, opened 2018-06-28
KG - Fix Viewport Width

kyle-glick:
The `no-margin` class was missing from catalog manager and causing the header to overflow to the side. Moving the class to the general layout sass file fixes the issue.


Stuart-Johnson:
@kyle-glick This was on my list of stuff, and it was driving me nutty! Thanks!

kyle-glick:
@Stuart-Johnson 

I can't scroll in peace when it randomly overflows to the right!

PR #1417, opened 2018-07-11
Short interactions [#158607645]

ying-pbrc:
PT: https://www.pivotaltracker.com/n/projects/1918597/stories/158607645


app/views/short_interactions/_form.html.haml
```diff
@@ -0,0 +1,45 @@
+-# Copyright © 2011-2016 MUSC Foundation for Research Development~
```
kyle-glick: Is this form partial being used? I can't seem to find where it would be rendered.

config/locales/proper.en.yml
```diff
@@ -285,3 +285,26 @@ en:
...
+      short_interaction:
+        header: "Short Interaction"
+        subject: "Subject of Interaction:*"
+        subject_select: " -- Select Category --"
+        type: "Type of Interaction:* "
+        type_select: "Select Type"
+        duration_placeholder: "Enter an integer"
+        duration_error: "Must enter an integer"
+        duration: "Duration in minutes:*"
+        email_error: "Must enter a valid email address"
```
kyle-glick: It looks like the various error messages here aren't being used. You can overwrite basic model errors using the `activerecord` hash in `en.yml`. We have a few examples of this already, but here's a link to the docs for this: http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models

app/views/service_requests/right_navigation/_short_interaction_modal.html.haml
```diff
@@ -0,0 +1,64 @@
...
+          %button.close{ type: 'button', data: { dismiss: 'modal' } }
+            %span{ aria: { hidden: 'true' } } &times;
+          %h4.text-center#modal-title
+            = t(:proper)[:right_navigation][:short_interaction][:header]
+        .modal-body
+          #modal_errors
+          .form-group.row
+            = f.label :subject, t(:proper)[:right_navigation][:short_interaction][:subject], class: 'col-lg-4 control-label'
+            .col-lg-8
+              = f.select :subject, options_for_select(PermissibleValue.get_inverted_hash('interaction_subject')), :prompt =>t(:proper)[:right_navigation][:short_interaction][:subject_select], class: 'form-control'
```
kyle-glick: From manual testing, it looks like this select isn't correctly applying the Bootstrap `form-control` class. Adding curly brackets around the `prompt` attribute fix the problem.

app/views/service_requests/right_navigation/_short_interaction_modal.html.haml
```diff
@@ -0,0 +1,64 @@
...
+          .form-group.row
+            = f.label :interaction_type, t(:proper)[:right_navigation][:short_interaction][:type], class: 'col-lg-4 control-label'
+            .col-lg-8
+              = f.select :interaction_type, options_for_select(PermissibleValue.get_inverted_hash('interaction_type')), :prompt =>t(:proper)[:right_navigation][:short_interaction][:type_select], class: 'form-control'
```
kyle-glick: See previous comment about the form selects. The same fix works here.

app/views/service_requests/right_navigation/_short_interaction_modal.html.haml
```diff
@@ -0,0 +1,64 @@
...
+            .col-lg-8
+              = f.text_field :name,  placeholder: t(:proper)[:right_navigation][:short_interaction][:investigator_name],  size: 65, class: 'form-control'
+          .form-group.row
+            = f.label :email, t(:proper)[:right_navigation][:short_interaction][:investigator_email], class: 'col-lg-4 control-label'
+            .col-lg-8
+              = f.text_field :email,  placeholder: t(:proper)[:right_navigation][:short_interaction][:investigator_email],  size: 65, class: 'form-control'
+          .form-group.row
+            = f.label :institution, t(:proper)[:right_navigation][:short_interaction][:institution], class: 'col-lg-4 control-label'
+            .col-lg-8
+              = f.select :institution, options_for_select(INSTITUTIONS), :prompt =>t(:proper)[:right_navigation][:short_interaction][:institution_select], class: 'form-control'
```
kyle-glick: See previous comment about the form selects. The same fix works here.

app/views/short_interactions/_form.html.haml
ying-pbrc: we had this for so long, it might be the original one that no longer needed. I will take a closer look

app/views/service_requests/right_navigation/_short_interaction_modal.html.haml
ying-pbrc: They are years old code! Thanks for catching. I will update them and do new push

PR #1422, opened 2018-07-13
Jl - Status sorting bug

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/156819921


kyle-glick:
I wonder if it would be more complete to apply this fix to the `PermissibleValue` model in general? Maybe as a default scope or just in the class methods?

PR #1442, opened 2018-07-27
wth - (SPARCDashboard, SPARCRequest & Data) Admin Rate Label Change a…

wtholt:
…nd Visual Cue

Background: Currently, the admin rate is an attribute on Line Item level. When an admin user set an admin rate on a service in a request, and then add the same service in SPARCFulfillment (which was not pushed from SPARCRequest), SPARCFulfillment considers it as a service with the normal rate rather than admin rate.

Please:

1). Make the format of the "Non-clinical Service" "Your Cost" editable field consistent with the other ones in SPARCDashboard Admin Edit Section “Clinical Services” Tab (see 2nd screenshot attached);

2). Add a visual cue (such as "(Modified)" sign or "M" icon with "Modified Rate" tooltip next to the cost in green color) for admin rate (SPARCRequest & SPARCDashboard).

The corresponding SPARCFulfillment story can be found here: https://www.pivotaltracker.com/story/show/156286625

[#156275541]

Story - https://www.pivotaltracker.com/story/show/156275541

### DIFF
--- a/app/assets/javascripts/dashboard/study_level_activities.js.coffee
+++ b/app/assets/javascripts/dashboard/study_level_activities.js.coffee
@@ -47,15 +47,6 @@ $ ->
       url: "/dashboard/line_items/#{line_item_id}/edit"
       data: data
 
-  $(document).on 'change', '.your-cost-edit', ->
-    row_index   = $(this).parents('tr').data('index')
-    line_item_id = $(this).parents('table.study_level_activities').bootstrapTable('getData')[row_index].id
-    data = 'line_item': 'displayed_cost' : $(this).val().replace('$', '')
-    $.ajax
-      type: 'PATCH'
-      url: "/dashboard/line_items/#{line_item_id}"
-      data: data
-
   $(document).on 'click', '.otf_delete', ->
     row_index   = $(this).parents('tr').data('index')
     line_item_id = $(this).parents('table.study_level_activities').bootstrapTable('getData')[row_index].id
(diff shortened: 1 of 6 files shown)


app/helpers/dashboard/study_level_activities_helper.rb
```diff
@@ -41,7 +41,9 @@ def sla_service_rate_display line_item
   end
 
   def sla_your_cost_field line_item
-    raw( text_field(:line_item, :service, class: 'your-cost-edit', value: number_with_precision(Service.cents_to_dollars(line_item.applicable_rate), precision: 2)) )
+    link_to 'javascript:void(0);', class: [("your-cost editable"), ('text-danger' if line_item.admin_rates.present?)], data: { url: dashboard_line_item_path(line_item), title: line_item.admin_rates.present? ? 'Modified Rate' : 'Admin Rate' }do
```
kyle-glick: Can you add this text to `en.yml`?

app/assets/javascripts/dashboard/study_level_activities.js.coffee
```diff
@@ -105,3 +96,12 @@ $ ->
       $.ajax
         type: 'DELETE'
         url: "/dashboard/fulfillments/#{fulfillment_id}"
+
+  $(document).on 'load-success.bs.table', '.study_level_activities', ->
+    setup_xeditable_fields()
+
+  $('body').tooltip(
```
kyle-glick: This is a good idea. At some point we should streamline our tooltips to do this using a global JS function.

wtholt: Yeah. The only reason I came across this was because tooltips and x-editable fields don't play nicely. Reason being the `data: { toggle: 'tooltip' }` part

app/helpers/dashboard/study_level_activities_helper.rb
wtholt: 👍

PR #1463, opened 2018-08-06
Short interaction institution migration [#159531739]

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/159531739

The institutions used in the "Short Interaction" function is currently pulling from constants.yml.
Please 1). migrate that to be using the institutions from professional_organizations in the database instead, and 2)remove the old attributes from constants.yml.


ying-pbrc:
I will clean up more and cover what Kyle has done in  https://github.com/sparc-request/sparc-request/pull/1458/files

PR #1479, opened 2018-08-13
ml - Past Status Data Cleanup

marklohr:
In the past_statuses table, there are a lot of historical records that have missing underscores.

The past statuses table should be saving the "machine name" of a status, not the "humanized name."

For example, "on_hold" is showing as "on hold", "in _process" is showing as "in process" etc, which was causing reporting issues.

Please write a script to clean up the historical data.


marklohr:
Run:    'rake data:clean_up_past_statuses'

PR #1564, opened 2018-09-21
KG - UDAK # (COEUS Project Number) Changes

kayla-glick:
https://www.pivotaltracker.com/story/show/157642975

It was requested to remove `udak` from the project number field, so I changed the YML text for the forms as well as renamed the field in the database.

In addition, the project number should be pulled from RMID (if enabled) and pre-populated / field disabled similar to all other RMID-enabled fields.


kyle-glick:
@Stuart-Johnson 

Hold on this PR for discussion about what to do with UDAK.

marklohr (approved):

PR #1604, opened 2018-10-29
Config Improvement [#161554164]

ying-pbrc:
PT: https://www.pivotaltracker.com/n/projects/1918597/stories/161554164
Scan for http security headers
https://securityheaders.com/

### DIFF
--- a/app/lib/settings_populator.rb
+++ b/app/lib/settings_populator.rb
@@ -27,9 +27,14 @@ class SettingsPopulator
 
   def populate
     ActiveRecord::Base.transaction do
+      puts "#{@defaults.size}"
+      updated = 0
+      inserted = 0
       @defaults.each do |hash|
         if Setting.exists?(key: hash['key'])
           Setting.find_by_key(hash['key']).update_attributes(hash.without('key', 'value'))
+          updated += 1
+          puts "#{updated}: #{hash['key']} updated" 
         else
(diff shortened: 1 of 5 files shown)


app/lib/settings_populator.rb
```diff
@@ -27,9 +27,14 @@ def initialize
...
   def populate
     ActiveRecord::Base.transaction do
+      puts "#{@defaults.size}"
+      updated = 0
+      inserted = 0
       @defaults.each do |hash|
         if Setting.exists?(key: hash['key'])
           Setting.find_by_key(hash['key']).update_attributes(hash.without('key', 'value'))
+          updated += 1
+          puts "#{updated}: #{hash['key']} updated"
```
kyle-glick: We originally had puts statements here to show the changes being made to Settings but it was requested to remove them, especially for the test suite, because it cluttered up spec feedback. If we put a condition on them so that they don't run when running tests then I think that should be fine.

ying-pbrc: Let me try to make that change and see if it pass travis

ying-pbrc: oops, didn't see that you have already made that change.

ying-pbrc:
my bad, i accidentally committed too many local changes,that i didn't mean to. Let me close this one and will resubmit my pull request.

PR #1627, opened 2018-11-13
ml-changes_to_Catalog_verbiage

marklohr:
In SPARCCatalog on all levels of User Rights, it says "Edit Historic Data" where it should say "Edit Historical Data" instead.


Stuart-Johnson:
https://www.pivotaltracker.com/story/show/161721921

PR #1637, opened 2018-11-28
JM - (SPARCRequest Step 2 & SPARCDashboard) Study Information Switching Funding Status Bug

jwiel86:
https://www.pivotaltracker.com/story/show/162136811

This story covers 2 funding source/status bugs described in above story


app/assets/javascripts/protocol_form.js.coffee
```diff
@@ -127,14 +127,30 @@ $(document).ready ->
...
     $('.funding_status_dependent').hide()
-    switch $(this).val()
-      when 'funded' then $('.funded').show()
-      when 'pending_funding' then $('.pending_funding').show()
-  ###END FUNDING STATUS FIELDS DISPLAY###
+    status_value = $(this).val()
 
+    if status_value == 'funded'
+      $('.funded').show()
+      source_value = $('#protocol_funding_source').val()
```
kyle-glick: You may need to initialize `source_value` to null outside of this conditional. In the case that the user selects the empty option (`status_value == ''`) then `source_value` will be uninitialized and will error out at the other conditional below, when really it should trigger the `else` block in that scenario

jwiel86: @kyle-glick Hey good eye! But that's a feature of javascript called hoisting. Basically your variables are scoped to the function they're defined in instead of the block they're defined in. So when you define a variable in a conditional block like that, it's declaration is actually "hoisted" to the top of the function but it won't actually be assigned a value until within the block where you actually defined it. So if it never reaches that block where it's assigned a value it'll just resolve to undefined which is falsy so it'll go into the else block.

PR #1638, opened 2018-11-28
AC / KG - Epic User Validation

kyle-glick:
https://www.pivotaltracker.com/story/show/162209368


amcates:
Kyle, this almost works.  If 'validate_epic_users' is set to 'false' it won't let me change the 'Epic EMR Access?' which it should.  Only if it's set to 'true' should the logic for disabled come in to play.  Otherwise it should always be enabled.  I will try to make a quick fix

app/models/epic_user.rb
```diff
@@ -0,0 +1,19 @@
...
+  def self.collection_name
+    @collection_name ||= Setting.find_by_key('epic_user_collection_name').try(:value)
+  end
+
+  def self.for_identity(identity)
+    get(:viewuser, userid: identity.ldap_uid.split('@').first)
+  end
+
+  def self.is_active?(epic_user)
+    epic_user.present? ? epic_user.key?('IsActive') && epic_user['IsActive'] : true
```
amcates: This fixes the use case where validate_epic_users is turned off (false).  Not sure if this is the correct way to handle it or not.  Up for suggestions

PR #1671, opened 2019-01-02
KG - Pricing Setup Labels

kyle-glick:
https://www.pivotaltracker.com/story/show/162506361

Pulls the labels on the Pricing Setup form from the `permissible_values` table.

Because I added the default scope ordering permissible values by `sort_order`, I removed all old code that did the same thing in various locations.

I also updated db/schema.rb because while trying to run specs, I found that `rake db:test:prepare` did not provide the correct database information


kyle-glick:
**Needs to be rebased**

PR #1719, opened 2019-02-12
Attempt to fix the affiliate netid issue

amcates:


amcates:
I tested this on -d and it works.  This is ready to be merged in order.  I  contains another capistrano pull request which is minor

PR #1720, opened 2019-02-13
Adding delayed job monitoring [#163951193]

amcates:


amcates:
FYI, the work in app/lib/directory just gets rid of the constant "directory.rb:52: warning: previous definition of LDAP_EMAIL was here".  We had them as constants but that is not what we needed.  Otherwise, any time you call a rake task or initialize the application they get re-instated

amcates:
Ah, specs are failing because the constant is missing.  I will fix and re-upload

amcates:
Specs are passing and I believe I fixed the duplicate constant declaration issue as well.

PR #1744, opened 2019-03-05
SRY - SPARCRequest confirmation page changes

royyesud:
Story: https://www.pivotaltracker.com/story/show/164176720

### DIFF
--- a/app/helpers/service_requests_helper.rb
+++ b/app/helpers/service_requests_helper.rb
@@ -77,14 +77,12 @@ module ServiceRequestsHelper
       remote: true, class: 'btn btn-default'
   end
 
-  def step_nav_button(text, color, url, inactive_link)
-    if !inactive_link
-      link_to(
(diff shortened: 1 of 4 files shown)


app/helpers/service_requests_helper.rb
```diff
@@ -78,13 +78,7 @@ def save_as_draft_button
...
 
   def step_nav_button(text, color, url, inactive_link)
-    if inactive_link
-      content_tag(:div,
-        content_tag(:div, raw(text), class: "btn step-text")+
-        content_tag(:div, '', class: "right-arrow"),
-        class: "step-btn step-btn-#{color} disabled_steps"
-      )
-    else
+    if !inactive_link
```
kyle-glick: The `inactive_link` logic can be removed entirely because it only exists to disable the step links on the confirmation page. It's used in `confirmation.html.haml` and `_steps.html.haml`

PR #1774, opened 2019-03-20
Jl - clincard split notify

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/164283835


lib/tasks/clinicard_process_ssrs_move.rake
```diff
@@ -0,0 +1,85 @@
+task clincard_process_ssrs_move: :environment do
```
kyle-glick: Is this something that can be made into a generalized task where we pass the organization ID to it? This task will only work for an MUSC instance of SPARCRequest and writing a new task for making changes to an organization individually seems redundant when we could have a single task that can be re-used.

jleonardw9: I'll rename it something more generic. You're right, it can be used for other organizations as well. I'll refactor to accept the id as an argument.

PR #1780, opened 2019-03-21
Fix Dashboard filter protocols bug

ying-pbrc:
See PT: 
https://www.pivotaltracker.com/n/projects/1918597/stories/164781694/comments/200780812

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -65,8 +65,7 @@ class Organization < ApplicationRecord
         where("super_users.identity_id = ? or service_providers.identity_id = ?", identity_id, identity_id).
         references(:super_users, :service_providers).
         distinct(:organizations).ids
-      ),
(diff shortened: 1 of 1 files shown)


app/models/organization.rb
```diff
@@ -65,8 +65,7 @@ class Organization < ApplicationRecord
         where("super_users.identity_id = ? or service_providers.identity_id = ?", identity_id, identity_id).
         references(:super_users, :service_providers).
         distinct(:organizations).ids
-      ),
```
kyle-glick: Should there also be a distinction made in the filter's dropdown that denotes inactive organizations?

ying-pbrc:
See
https://github.com/sparc-request/sparc-request/blob/master/app/lib/dashboard/grouped_organizations.rb#L45

PR #1831, opened 2019-04-17
Fix add/remove ssrs bug when protocol has not yet created

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/165421356/comments/201722625

### DIFF
--- a/app/models/service_request.rb
+++ b/app/models/service_request.rb
@@ -437,12 +437,10 @@ class ServiceRequest < ApplicationRecord
     next_ssr_id = self.protocol && self.protocol.next_ssr_id.present? ? self.protocol.next_ssr_id : 1
 
     self.sub_service_requests.each do |ssr|
-      unless ssr.ssr_id
+      unless ssr.ssr_id && self.protocol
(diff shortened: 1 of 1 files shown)


app/models/service_request.rb
```diff
@@ -437,12 +437,10 @@ def ensure_ssr_ids
     next_ssr_id = self.protocol && self.protocol.next_ssr_id.present? ? self.protocol.next_ssr_id : 1
 
     self.sub_service_requests.each do |ssr|
-      unless ssr.ssr_id
+      unless ssr.ssr_id && self.protocol
```
kyle-glick: @ying-pbrc Can you explain what was, from a technical perspective, causing this issue? The way I'm reading the code sounds to me like this would prevent Sub Service Requests from properly assigning an `ssr_id` if there is no protocol yet?

ying-pbrc: While it only updates the ssr_id for the Sub Service Requests that do not have an assigned ssr_id if protocol exist, it should reorder/update all SSRs in the correct order 0001, 0002, 0003, ... etc if there in not yet created.

PR #1876, opened 2019-06-11
All requested services should be displayed on Step 2

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/166545762


ying-pbrc:
it created another issue on Step 1. I will rework on this

PR #1904, opened 2019-06-27
Archive protocol email bug

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/166791970

In current release (V3.5.0) only one email was sent to one of the intended recipients when someone archive/unarchive a protocol on Dashboard. This was because a single Action Mailer call can only render one view. This issue can be resolved by moving the recipient list from protocol_mailer to dashboard protocol controller and calling the protocol mailer for each recipient.


config/locales/en.yml
```diff
@@ -426,7 +426,7 @@ en:
           archiver: "(Archiver)"
         unarchive:
           subject: "SPARCRequest Protocol Unarchive Notification (Protocol %{protocol_id})"
-          introduction: "Protocol %{protocol_id} has been unarchived in SPARCRequest by %{archiver_name}. Your protocol can still be accessed through archived protocols or by <a href='%{link}' target='blank'>this direct link to SPARCDashboard</a>."
+          introduction: "Protocol %{protocol_id} has been unarchived in SPARCRequest by %{archiver_name}. Your protocol can be accessed on your SPARCDashboard or by <a href='%{link}' target='blank'>this direct link to SPARCDashboard</a>."
```
kayla-glick: My only suggestion is related to the language of the introduction. Right now it's

> Protocol %{protocol_id} has been unarchived in SPARCRequest by %{archiver_name}. Your protocol can be accessed on your SPARCDashboard or by this direct link to SPARCDashboard

- We should replace `unarchived` with the action from the mailer (`archived`/`unarchived`)
- `to SPARCDashboard` now sounds redundant to me with the added `on your SPARCDashboard`. Should we simply remove that part so that the link says `this direct link`?

ying-pbrc: I have updated the language of the unarchived introduction as follows:

-  Protocol %{protocol_id} has been unarchived in SPARCRequest by %{archiver_name}. Your protocol can be accessed by this direct link to SPARCDashboard.

However, I think the language of the archived introduction should remain as is. 
-  Protocol %{protocol_id} has been archived in SPARCRequest by %{archiver_name}. Your protocol can still be accessed through archived protocols or by this direct link to SPARCDashboard.

It helps to remind users they to check the Archived protocol filter chceckbox first in order to see their archived protocols in SPARCDashboard.

kayla-glick: @ying-pbrc

Looks good to me. Disregard my first suggestion. I just realized we're already distinguishing the text for archiving vs unarchiving.

PR #1935, opened 2019-07-17
Patch 1

wsams:
It's been useful to support environment variables in our instance of sparc-request. This change allows environment variable substitution in `fulfillment_db.yml`. I have a similar change for `sparc_db.yml` in sparc-fulfillment. Will submit shortly. Thanks.


wsams:
I'm sorry, I messed up this commit. Should only have been one file changed. I'll resubmit.

PR #1949, opened 2019-07-30
Sparc user profile

ying-pbrc:
New edit sparc user profile using existing route from devise gem, in bootstrap 4. 
See https://www.pivotaltracker.com/n/projects/1918597/stories/166341861


app/views/devise/registrations/_profile_form.html.haml
```diff
@@ -0,0 +1,58 @@
+= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
+  = render "devise/shared/error_messages", resource: resource
```
kayla-glick: This will place all of the errors in an alert at the top of the form. Can you take a look at the logic for the new user sign up form that places the errors messages on their corresponding field?

app/views/devise/shared/_error_messages.haml
```diff
@@ -0,0 +1,5 @@
+- if resource.errors.any?
```
kayla-glick: See previous comment. Use inline errors instead of block errors

config/locales/devise.en.yml
```diff
@@ -126,3 +126,24 @@ en:
         subject: "SPARCRequest Reset Password Instructions"
       unlock_instructions:
         subject: "SPARCRequest Unlock Instructions"
+    ###########
+    # Profile #
+    ###########
+    profile:
+      header: "User Profile"
+      update: "Save Changes"
+      full_name: "Full Name"
```
kayla-glick: Most of these attributes are already available via the ActiveRecord YAML (see https://stackoverflow.com/questions/10812774/how-can-i-translate-the-activerecord-attribute-name-in-rails-3-2). Using `f.label :attribute_name` will pull from those existing YAML fields. See the Authorized User form and new user sign up form.

PR #1957, opened 2019-08-05
Fixing issue with delayed job not knowing current user on model save

amcates:


kayla-glick:
Can you share the error message that this is related to?

sparc-request:
It’s not an error message.  The issue is since it’s a delayed job, current_user doesn’t exist anymore.  I have this fixed.

PR #1972, opened 2019-08-19
Bbootstrap 4 - SPARCFunding

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/166643702


app/views/funding/services/show.html.haml
```diff
@@ -18,14 +18,17 @@
...
+%header.d-flex.mb-2.default
+  %h2.p-2.mr-auto
+    = t('funding.download.header')
+  %h4.p-2
+    = link_to "/funding?back=true" do
+      = icon('fas', 'arrow-left mr-2')
+      = t('funding.download.back')
+
+.card
+  .card-header.bg-default.text-dark.h4
```
kayla-glick: Can you also change this header like the other one?

app/views/funding/services/show.html.haml
```diff
@@ -18,14 +18,17 @@
...
     = render 'funding/services/title', service: @service
+  .card-body.mt-3
```
kayla-glick: This card/table I'm not entirely sure about. Can you try something and let me know how it looks? To make the table consistent like on the other card, can you move it outside the `.card-body` so that it's directly under the `.card`? Then can you wrap the two buttons inside a `.card-body` so that they have the padding we want? That will probably put a line between the buttons and the table, but I'm wondering if you add a `.border-bottom-0` to the `.card-body` if that would look okay?

app/views/funding/services/_documents.html.haml
```diff
@@ -18,41 +18,34 @@
...
-          = t(:funding)[:download][:table][:short_title]
-        %th{data: {class: 'col-sm-3 doc_title', align: "left", sortable: "false", field: "doc_title"}}
-          = t(:funding)[:download][:table][:doc_title]
-        %th{data: {class: 'col-sm-1 uploaded', align: "left", sortable: "true", field: "uploaded", sorter: "dateSorter"}}
-          = t(:funding)[:download][:table][:uploaded]
-        %th{data: {class: 'col-sm-1 status', align: "left", sortable: "true", field: "status"}}
-          = t(:funding)[:download][:table][:status]
-        %th{data: {class: 'col-sm-2 Actions', align: "center", sortable: "false", field: "actions"}}
-          = t(:funding)[:download][:table][:actions][:label]
+.row.p-3
```
kayla-glick: Could you remove this `.row` and `.col-lg-12`? They don't really serve a purpose since there's only one element

app/views/funding/services/show.html.haml
```diff
@@ -18,14 +18,17 @@
...
-    = link_to "/funding?back=true", title: "Back to funding opportunities" do
-      %span.glyphicon.glyphicon-arrow-left{aria: {hidden: "true"}}
-      = t(:actions)[:back]
-  .panel-body
+
+%header.d-flex.mb-2.default
+  %h2.p-2.mr-auto
+    = t('funding.download.header')
+  %h4.p-2
+    = link_to "/funding?back=true" do
```
kayla-glick: Is the `back` param actually used here?

app/views/funding/services/show.html.haml
```diff
@@ -18,14 +18,17 @@
...
+      = icon('fas', 'arrow-left mr-2')
+      = t('funding.download.back')
```
kayla-glick: When using content inside a `link_to ... do` like this HAML doesn't trim whitespace which causes extra spacing between elements. You can fix that by doing

```
- succeed <text> do
  = <content before text>
```

app/views/funding/services/index.html.haml
```diff
@@ -18,20 +18,19 @@
...
-            %th.col-md-2{ data: { field: "action", align: "center" } }
-              = t(:funding)[:index][:table][:headers][:action]
\ No newline at end of file
+.card#fundingsList
+  .card-header.bg-default.text-dark
+    %h2.p-1
+      = t('funding.index.header')
+  .card-body
+    %table#fundings_table{ data: { toggle: 'table', url: funding_services_path(format: :json), search: 'true', 'show-columns' => 'true', 'show-refresh' => 'true', 'show-export' => 'true', pagination: 'true', page_size: '20', sort_name: 'created', sort_order: 'desc', toolbar: '#fundings-custom-toolbar' } }
+      %thead.bg-primary.text-white
```
kayla-glick: Can you make this `thead.bg-light` to make it consistent with other tables?

app/views/funding/services/_documents.html.haml
```diff
@@ -18,41 +18,34 @@
...
-      .col-lg-6
-        %button.btn.btn-default.btn-block.document_nav#documents_application
-          %span.pull-left
-            %span.glyphicon.glyphicon-file
-            %span
-              = t(:funding)[:download][:application]
-          %span.glyphicon.glyphicon-refresh.pull-right{style: "display: none;"}
+.row.mb-3#documents_tabs{data: {selected: "loi"}}
+  .col-lg-6
+    %button.btn.btn-primary.btn-lg.btn-block.document_nav#documents_loi
```
kayla-glick: Adding a `<` character at the end of this line (no space between) will remove the extra whitespace between the icon and text

app/views/funding/services/_documents.html.haml
```diff
@@ -18,41 +18,34 @@
...
+      = icon('fas', 'file mr-2')
+      = t(:funding)[:download][:loi]
+  .col-lg-6
+    %button.btn.btn-light.btn-lg.btn-block.document_nav#documents_application
```
kayla-glick: See previous comment

app/views/funding/services/show.html.haml
ying-pbrc: I moved the table outside the .card-body and placed it directly under the .card, It looks okay
image

kayla-glick: @ying-pbrc Are you sure the table isn't still inside of the `.card-body`? It looks like it is from that screenshot because of the padding around the sides of the table

ying-pbrc: updated screenshot, just pushed the new commit to upstream
image

PR #2005, opened 2019-09-10
fixing small typo

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/168408076

### DIFF
--- a/app/controllers/dashboard/associated_users_controller.rb
+++ b/app/controllers/dashboard/associated_users_controller.rb
@@ -81,7 +81,7 @@ class Dashboard::AssociatedUsersController < Dashboard::BaseController
     if updater.successful?
       flash[:success] = t('authorized_users.updated')
 
-      redirect_to dashboard_root_path if @protocol_role.identity == current_user && !@admin && ['none'].include?(protocol_role.project_rights)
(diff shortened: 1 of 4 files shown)


kayla-glick:
@ying-pbrc Thanks for fixing this! I also noticed a similar bug today when I tried to edit a user. I think the scenario is that Epic is turned on and the protocol is selected for epic:

```
An ActionView::Template::Error occurred in associated_users#edit:

  undefined local variable or method `epic_user' for #<#<Class:0x00007ff8c7361678>:0x00007ff8c735b638>
Did you mean?  @epic_user
  app/views/associated_users/_user_form.html.haml:108:in `block in _app_views_associated_users__user_form_html_haml___1772019914063096418_70353235353360'
```

Looking at the code, I think we just need to pass `epic_user: @epic_user` when the template renders the `user_form` partial. Would you like to look into that and add it to this PR or should I make a separate pull request?

ying-pbrc:
I can take a look at it. We have turned epic off so I am not too familiar with epic related functionality. Does it only occur in Dashboard? or in SPARC Request Step 2 also?

kayla-glick:
@ying-pbrc It occurs in both places

ying-pbrc:
@kayla-glick . Okay, I will take a look.

PR #2016, opened 2019-09-24
Jl - service creation bug

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/167112401


kayla-glick:
I added the `available` scope to those dropdowns for a previous story to prevent users from moving an active service to an inactive organization. Would it be possible to do a conditional statement like if the service is inactive then show inactive organizations otherwise don't?

jleonardw9:
@kayla-glick Added a new scope and made both dropdowns into conditionals.

PR #2023, opened 2019-09-30
SAW - Bootstrap 4 Survey Builder and SPARCForms

SamWord:
Updated SPARCForms and the Survey Builder to use Bootstrap 4 [#166643537](https://www.pivotaltracker.com/story/show/166643537)

This pull request also addresses several other survey bugs:

- Text area length bug where some survey response fields only allow up to 255 characters despite having no limit in the database [#168031586](https://www.pivotaltracker.com/story/show/168031586)
- Multiple Choice with Dependent Question where certain dependent questions produce an error when viewing or editing the response [#168031795](https://www.pivotaltracker.com/story/show/168031795)
- Form builder not displaying actions after using the search bar [#168031921](https://www.pivotaltracker.com/story/show/168031921)
- Dashboard error when there are incomplete forms on a service request [#168699471](https://www.pivotaltracker.com/story/show/168699471)
- Requests table on the Dashboard disappears after editing an authorized user (this one wasn't related to surveys, but an identical issue was found with the forms table in Dashboard so I fixed this one as well) [#168741305](https://www.pivotaltracker.com/story/show/168741305)


kayla-glick:
These are the only Issues I found while testing, otherwise looks good to me:

- When removing a response filter the loading bar keeps spinning. It just needs an `ajax:complete` trigger
- The alignment is a little off for radio button and checkbox options 
image
- When creating a new survey/form the loading bar keeps spinning. It just needs an `ajax:complete` trigger

PR #2054, opened 2019-10-31
Edit profile refactor

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/166341861/comments/208192748

-  Added a 'noneditable_identity_fields' setting to determine whether or not the first name, last name, and email fields on Edit Profile page are editable or not.
- Make first name, last name, and email fields read-only based on the 'noneditable_identity_fields' setting.
- Restructure the edit profile form to be more in sync with the new user form for consistency.


db/migrate/20191031190917_add_noneditable_identity_fields_setting.rb
```diff
@@ -0,0 +1,15 @@
+class AddNoneditableIdentityFieldsSetting < ActiveRecord::Migration[5.2]
```
kayla-glick: Rather than adding this setting via migration, can you add it to `defaults.json`? It can then be imported via the rake task. I think we already have settings changes anyways for this release

kayla-glick: Everything else looks good!

ying-pbrc:
Kayla, please see my reply in red below

db/migrate/20191031190917_add_noneditable_identity_fields_setting.rb
kayla-glick: @ying-pbrc can you remove this migration? That way we can make sure that the setting created matches what's listed in `defaults.json`. Migrations are really meant to be for table/column changes rather than creating data.

PR #2056, opened 2019-11-01
SJ - Sparc Cloud Branch

Stuart-Johnson:
Listing out stories below:

[#168904200]


db/migrate/20191101150844_add_ontology_tag_system.rb
```diff
@@ -0,0 +1,29 @@
+class AddOntologyTagSystem < ActiveRecord::Migration[5.2]
+  def change
+    ## Add column to store ontology tag on organization
+    add_column :organizations, :ontology_tag, :string
+
+    ## Populate new ontology tag permissible values
+    unless PermissibleValue.where(category: 'ontology_tag').exists?
```
kayla-glick: Instead of creating the PermissibleValues here, can you add a source CSV file and run the import task?

kayla-glick (dismissed):

amcates:
@jleonardw9 @Stuart-Johnson I believe this pull request should just be trashed.  It's really old and I think we have moved past this.  Thoughts?

Stuart-Johnson:
@amcates Agreed, I was just never sure if this was something that needed to be revisited.

PR #2064, opened 2019-11-12
SJ - Fixing service search when not logged in 

Stuart-Johnson:
Request format for services search was json, and code was only allowing js past the check.

[#169708113]

https://www.pivotaltracker.com/story/show/169708113

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -171,7 +171,7 @@ class ApplicationController < ActionController::Base
 
   def authorize_identity
     # If the request is in first_draft status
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -171,7 +171,7 @@ def initialize_service_request
 
   def authorize_identity
     # If the request is in first_draft status
-    if @service_request.status == 'first_draft' && (action_name == 'catalog' || (Rails.application.routes.recognize_path(request.referrer)[:action] == 'catalog' && request.format.js?))
+    if @service_request.status == 'first_draft' && (action_name == 'catalog' || (Rails.application.routes.recognize_path(request.referrer)[:action] == 'catalog' && (request.format.js? || request.format.json?)))
```
kayla-glick: Actually I'm wondering if it would be better to do `!request.format.html?` to cover all cases?

PR #2123, opened 2019-12-17
SJ - Fix for recognize path issues

Stuart-Johnson:
This is a somewhat temporary fix for various recognize path issues. Recognize path is VERY deprecated, so we will want to rework all these locations for next release anyways. However, in the meantime, we want to handle issues more gracefully (not throw 500 errors)

### DIFF
--- a/app/controllers/surveyor/responses_controller.rb
+++ b/app/controllers/surveyor/responses_controller.rb
@@ -95,7 +95,7 @@ class Surveyor::ResponsesController < Surveyor::BaseController
     @permission_to_edit = @protocol_role.nil? ? false : @protocol_role.can_edit? if @protocol
 
     if @response.save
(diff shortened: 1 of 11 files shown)


app/controllers/surveyor/responses_controller.rb
```diff
@@ -95,7 +95,8 @@ def create
     @permission_to_edit = @protocol_role.nil? ? false : @protocol_role.can_edit? if @protocol
 
     if @response.save
-      SurveyNotification.system_satisfaction_survey(@response).deliver_now if @response.survey.access_code == 'system-satisfaction-survey' && Rails.application.routes.recognize_path(request.referrer)[:action] == 'review'
+      action = Rails.application.routes.recognize_path(request.referrer)[:action] rescue nil
+      SurveyNotification.system_satisfaction_survey(@response).deliver_now if @response.survey.access_code == 'system-satisfaction-survey' && action == 'review'
```
kayla-glick: Let's change this to just use the `in_review?` helper to minimize our usage of `#recognize_path`

PR #2142, opened 2019-12-19
KG - Improve Calendar Check/Uncheck Row/Column

kayla-glick:
https://www.pivotaltracker.com/story/show/170387345

There are actually two issues here:
1. `update_all` skips callbacks, meaning none of the updates from toggling the rows or columns were being audited
2. `update_all` is significantly slower on smaller numbers of records (< 200 or so). While normally using n `update_attribute` calls would be expensive, wrapping it in a transaction significantly improves performance at larger numbers of objects while only barely decreasing performance at smaller numbers

### DIFF
--- a/app/controllers/service_calendars_controller.rb
+++ b/app/controllers/service_calendars_controller.rb
@@ -122,7 +122,7 @@ class ServiceCalendarsController < ApplicationController
           v.update_attributes(quantity: unit_minimum, research_billing_qty: unit_minimum, insurance_billing_qty: 0, effort_billing_qty: 0)
         end
       elsif params[:uncheck]
-        @visits.each{ |v| v.update_attributees(quantity: 0, research_billing_qty: 0, insurance_billing_qty: 0, effort_billing_qty: 0) }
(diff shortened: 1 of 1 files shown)


app/controllers/service_calendars_controller.rb
```diff
@@ -114,14 +115,15 @@ def toggle_calendar_column
...
-        
-        v.update_attributes(quantity: unit_minimum, research_billing_qty: unit_minimum, insurance_billing_qty: 0, effort_billing_qty: 0)
+    Visit.transaction do
+      if params[:check]
+        @visits.each do |v|
+          unit_minimum = v.service.displayed_pricing_map.unit_minimum
+          v.update_attributes(quantity: unit_minimum, research_billing_qty: unit_minimum, insurance_billing_qty: 0, effort_billing_qty: 0)
+        end
+      elsif params[:uncheck]
+        @visits.each{ |v| v.update_attributees(quantity: 0, research_billing_qty: 0, insurance_billing_qty: 0, effort_billing_qty: 0) }
```
Stuart-Johnson: Got a typo here, I'll just fix it and re-push

PR #2165, opened 2020-01-13
Fix Cancel Edit Protocol redirect link from Dashboard 

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/170650706


app/views/protocols/form/_footer.html.haml
```diff
@@ -21,7 +21,7 @@
 %footer.card.w-100.sticky-bottom.mt-3#protocolFooter
   .card-body.py-2.px-5
     .form-group.mb-0<
-      - back_url = in_dashboard? ? dashboard_root_path : protocol_service_request_path(srid: @service_request.try(:id))
+      - back_url = in_dashboard? ? dashboard_protocol_path(@protocol) : protocol_service_request_path(srid: @service_request.try(:id))
```
kayla-glick: This needs an additional condition checking whether the user is creating a new protocol or editing an existing protocol. For a new protocol in should go to the dashboard root path and for editing it should go to the dashboard protocol path

ying-pbrc:
Yes, I just realized that, will fix it.

PR #2178, opened 2020-01-22
KG - Fix RMID Search & Study Type Note Authorization

kayla-glick:
https://www.pivotaltracker.com/story/show/170853298

Fixes permissions for general users when:
1. Searching for an RMID record on the Protocol form
2. Changing the study type answers and displaying a study type note on the Protocol form

### DIFF
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -21,9 +21,9 @@
 class ProtocolsController < ApplicationController
   respond_to :html, :js, :json
 
-  before_action :initialize_service_request,  except: [:approve_epic_rights, :push_to_epic, :push_to_epic_status]
-  before_action :authorize_identity,          except: [:approve_epic_rights, :push_to_epic, :push_to_epic_status]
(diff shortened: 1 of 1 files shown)


app/controllers/protocols_controller.rb
```diff
@@ -21,9 +21,9 @@
 class ProtocolsController < ApplicationController
   respond_to :html, :js, :json
 
-  before_action :initialize_service_request,  except: [:approve_epic_rights, :push_to_epic, :push_to_epic_status]
-  before_action :authorize_identity,          except: [:approve_epic_rights, :push_to_epic, :push_to_epic_status]
-  before_action :find_protocol,               only:   [:edit, :update, :show]
+  before_action :initialize_service_request,  only: [:show, :new, :create, :edit, :update, :update_protocol_type]
```
kayla-glick: The `validate_rmid` and `get_study_type_note` routes were running initialize_service_request and authorize_identity when they shouldn't have been. I changed the except to and only to prevent that from happening by mistake in the future with new routes

PR #2205, opened 2020-01-30
Jl - Historical protocol merges

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/168226013

### DIFF
--- a/app/controllers/dashboard/protocols_controller.rb
+++ b/app/controllers/dashboard/protocols_controller.rb
@@ -33,6 +33,16 @@ class Dashboard::ProtocolsController < Dashboard::BaseController
 
     @default_filter_params  = { show_archived: 0 }
 
+    # if we are performing a search, check if user is looking for an old protocol
+    # that has been merged and return the most current master protocol
+    if params.has_key?(:filterrific) && params[:filterrific].has_key?(:search_query)
+      search_term = params[:filterrific][:search_query][:search_text].to_i
+      merge = search_protocol_merges(search_term)
+      if merge
+        params[:filterrific][:search_query][:search_text] = merge.master_protocol_id.to_s
+      end 
+    end
+
     # if we are an admin we want to default to admin organizations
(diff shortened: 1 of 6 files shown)


app/controllers/dashboard/protocols_controller.rb
```diff
@@ -218,4 +228,17 @@ def filterrific_params
       with_status: [],
       with_owner: [])
   end
+
+  def search_protocol_merges(protocol_id)
```
kayla-glick: Can you give a short explanation of what you added in this controller? I'm struggling to follow what it's doing. Thanks!

app/helpers/dashboard/protocols_helper.rb
```diff
@@ -64,4 +64,14 @@ def display_requests_button(protocol, access)
       end
     end
   end
+
+  def display_merges(protocol)
```
kayla-glick: I think we should probably add a `has_many :protocol_merges` association to the Protocol model then you can replace this helper method with `protocol.protocol_merges.pluck(:merged_protocol_id)` which will return the merged protocol IDs. Without the association it would be `ProtocolMerge.where(master_protocol: protocol).pluck(:merged_protocol_id)`

app/views/dashboard/protocols/_table.html.haml
```diff
@@ -40,6 +40,8 @@
         %tr
           %th.w-10{ data: { field: 'id', align: 'center', sortable: 'true' } }
             = Protocol.human_attribute_name(:id)
+          %th.w-10{ data: { field: 'protocol_merges', align: 'center', sortable: 'true', visible: 'false' } }
```
kayla-glick: Could you put the text in YAML (the attribute names are coming from `en.activerecord.models.attributes.<atttribute_name>` so you could define one called `merged_protocols: "Historical Protocol Merges"` or something along those lines

PR #2297, opened 2020-04-15
Jl - Protocol merge ssr ordering and new validations

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/169658017


app/controllers/dashboard/protocol_merges_controller.rb
```diff
@@ -159,4 +162,21 @@ def role_should_be_assigned?(role_to_be_assigned, protocol)
   def has_research?(protocol, research_type)
     protocol.research_types_info.try(research_type) || false
   end
+
+  # We can not merge if 2 requests are under the same process ssrs org AND either of those requests are locked or incomplete
+  def requests_not_merge_eligible?(master, merged)
+    master_requests = master.sub_service_requests
+    merged_requests = merged.sub_service_requests
+
+    master_requests.each do |master_request|
```
kayla-glick, line 171: I'm wondering if this can be made more efficient because it's currently O(n^2).

Based on the condition for whether two protocols can't be merged, I think we could do this
```
# Grab all locked / incomplete requests from each protocol
master_ineligible_requests = master.sub_service_requests.select{ |ssr| ssr.is_locked? || !ssr.is_complete?}
merged_ineligible_requests = merged.sub_service_requests.select{ |ssr| ssr.is_locked? || !ssr.is_complete?}

# Take the intersection of the potentially ineligible requests to see if any of them have the same "organization_id"
return (master_ineligible_requests.map(&:organization_id) & merged_ineligible_requests.map(&:organization_id)).any?
```

Which would be O(2n)

app/controllers/dashboard/protocol_merges_controller.rb
```diff
@@ -55,13 +55,16 @@ def perform_protocol_merge
...
+        @errors[:master_protocol_id] = t(:dashboard)[:protocol_merge][:errors][:split_notify_error]
       elsif @errors.empty? && !confirmed
         @no_errors = true
         return
       else
         ActiveRecord::Base.transaction do
+          merged_ssr_ids = @merged_protocol.sub_service_requests.map(&:id)
           merge_srs = Dashboard::MergeSrs.new()
-          fix_ssr_ids = Dashboard::FixSsrIds.new(@master_protocol)
+          fix_ssr_ids = Dashboard::FixSsrIds.new(@master_protocol, merged_ssr_ids)
```
kayla-glick, line 67: You could pass the `merged_ssrs` directly instead of the IDs to avoid having to re-find them in the `FixSsrIds` file

app/controllers/dashboard/protocol_merges_controller.rb
jleonardw9: I like it and will make the requested changes.

app/controllers/dashboard/protocol_merges_controller.rb
jleonardw9: I think I'm going to have to leave this one as is. I can explain more offline if you'd like, but it has to do with the ssr's protocol id changing with the merge.

PR #2308, opened 2020-04-22
SJ - Fix for phone format

Stuart-Johnson:
If the if statement is false, the last line returned is technically false, not the actual phone number. So all 10 or less digit phone numbers were just showing as nil (a dash, in the front end).

[#172447401]

https://www.pivotaltracker.com/story/show/172447401


kayla-glick:
@Stuart-Johnson I actually did this intentionally when I wrote the helper to try to force users to update phone numbers to match our new forced formatting

PR #2314, opened 2020-04-23
Fixes error trying to add new user with ldap uid

chrisortman:
If we're using ldap and lazy_ldap the suggestion value
is the `ldap_uid` and not the `id` but the query used find
which only looks at `id`


chrisortman:
I see this was fixed in 4b7cb521bf0e777a289d237c4f9a1b483c1108f8

PR #2362, opened 2020-05-20
Jl - Non-clinical line item fulfillment synch

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/172409825


Stuart-Johnson:
I'm merging this in because Wenjun wants it NOW for testing, and creating another PR to fix specs...

PR #2369, opened 2020-05-22
Jl - otf fulfillment line item synch bugs

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/172409825/comments/214617467

### DIFF
--- a/app/helpers/dashboard/study_level_activities_helper.rb
+++ b/app/helpers/dashboard/study_level_activities_helper.rb
@@ -32,7 +32,7 @@ module Dashboard::StudyLevelActivitiesHelper
   def sla_service_name_display(line_item)
     text      = line_item.service.display_service_name
     text     += inactive_tag unless line_item.service.is_available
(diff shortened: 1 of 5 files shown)


app/helpers/dashboard/study_level_activities_helper.rb
```diff
@@ -32,7 +32,7 @@ def sla_fulfillments_button(line_item)
   def sla_service_name_display(line_item)
     text      = line_item.service.display_service_name
     text     += inactive_tag unless line_item.service.is_available
-    text     += " (In Work Fulfillment)" if sla_in_fulfillment?(line_item)
+    text     += content_tag :span, style: 'color:green;' do " (In SPARCFulfillment)" if sla_in_fulfillment?(line_item) end
```
kayla-glick, line 35: We should use the `text-success` class here for consistency instead of inline styling

jleonardw9: Ok, changed to 'text-success'

PR #2396, opened 2020-06-05
Adds rake tasks for db backup/restore

chrisortman:
These are a couple of rake tasks I've found helpful for using in deployment or when troubleshooting.

Can be invoked with `rails mysql:backup` `rails mysql:restore`

There is also `rails mysql:dev_settings` you can use if you
are restoring a production backup to your development machine

### DIFF
--- a/lib/tasks/mysql.rake
+++ b/lib/tasks/mysql.rake
@@ -1,0 +1,70 @@
+namespace :mysql do
+  desc 'execute a SQL command. Usage: rake db:execute SQL=""'
+  task :execute => :environment do
+    ActiveRecord::Base.establish_connection( (ENV['ENV'] || Rails.env).to_sym)
+    ActiveRecord::Base.connection.execute( ENV['SQL'].to_s ).each { |hash| p hash }
+  end
+
+  desc 'Update values in settings table for development. Useful if you restore production database'
+  task :dev_settings => :environment do
+    Setting.set_value!( 'admin_mail_to','admin@example.com')
+    Setting.set_value!( 'clinical_work_fulfillment_url','http://localhost:4000')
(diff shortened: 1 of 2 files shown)


Stuart-Johnson:
Just a note here, this build passed Travis, the link just isn't showing up for some reason.

lib/tasks/mysql.rake
```diff
@@ -0,0 +1,70 @@
+namespace :mysql do
+  desc 'execute a SQL command. Usage: rake db:execute SQL=""'
+  task :execute => :environment do
+    ActiveRecord::Base.establish_connection( (ENV['ENV'] || Rails.env).to_sym)
+    ActiveRecord::Base.connection.execute( ENV['SQL'].to_s ).each { |hash| p hash }
+  end
+
+  desc 'Update values in settings table for development. Useful if you restore production database'
+  task :dev_settings => :environment do
```
kayla-glick, line 9: Would it potentially be a little cleaner to truncate the table and re-run the populator instead of manually assigning the values here? That way it's consistent with `defaults.json`

Stuart-Johnson:
@chrisortman I don't think we are going to use this one, as the "regenerate settings" rake task already covers some of this functionality, in a way that works better for our workflow.

lib/tasks/mysql.rake
chrisortman: Maybe, but I guess in my case I don't want the settings that are in defaults.json, I want _most_ of the settings we've configured in production.

Stuart-Johnson:
@chrisortman That's actually what an application.yml is for. That's what we use on our testing server. It will look there first for values, then use defaults if it can't find one.

PR #2434, opened 2020-07-22
Jl - Select all t quantities

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/165912000

### DIFF
--- a/app/assets/stylesheets/service_calendar.scss
+++ b/app/assets/stylesheets/service_calendar.scss
@@ -26,6 +26,7 @@
 }
 
 #serviceCalendar .service-calendar-container {
+
   &:not(:last-of-type) {
     @extend .border-bottom;
     border-bottom-left-radius: 0;
@@ -119,6 +120,10 @@
     }
   }
 }
+
+#select-t-qty {
+    margin-top: 10px;
+  }
 
 // Admin Styling
 .card #serviceCalendar {
--- a/app/assets/stylesheets/service_calendar.scss
+++ b/app/assets/stylesheets/service_calendar.scss
@@ -121,10 +121,6 @@
   }
 }
 
-#select-t-qty {
-    margin-top: 10px;
-  }
-
 // Admin Styling
 .card #serviceCalendar {
   nav .nav-link:first-of-type {
--- a/app/controllers/visits_controller.rb
+++ b/app/controllers/visits_controller.rb
@@ -42,11 +42,8 @@ class VisitsController < ApplicationController
 
     if @visit.update_attributes(visit_params)
       @visit.sub_service_request.set_to_draft unless @in_admin
-      if visit_params[:set_all_insurance] == '1'
-        insurance_qty = visit_params[:insurance_billing_qty].to_i
-        @visit_group.visits.each do |visit|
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/service_calendar.scss
```diff
@@ -120,6 +121,10 @@
   }
 }
 
+#select-t-qty {
```
kayla-glick, line 124: Instead of specifying the margin here, you can remove the `mb-0` from the 3 quantities `form-group`s which will then have a built-in `margin-bottom` for spacing with the stuff you added

app/views/visits/_form.html.haml
```diff
@@ -43,6 +43,9 @@
           .form-group.col-4.mb-0
             = f.label :effort_billing_qty, class: 'required'
             = f.text_field :effort_billing_qty, class: 'form-control'
+          .form-group.col-4{ id: 'select-t-qty' }
+            = f.label t(:calendars)[:pppv][:visits][:select_all_t]
```
kayla-glick, line 47: Can you set the `for` attribute (first parameter) on this label so clicking it will toggle the checkbox?

app/views/visits/_form.html.haml
```diff
@@ -43,6 +43,9 @@
...
+            = f.check_box :set_all_insurance
```
kayla-glick, line 48: Can you add the Bootstrap 4 styles for this checkbox for consistency? We can also make the checkbox a bit bigger so that it's more prominent by using the custom checkbox element styles (see the documentation below) and adding `custom-control-lg` to the `custom-control` element, which we do in a few places in the application for larger checkboxes.

https://getbootstrap.com/docs/4.5/components/forms/#checkboxes

app/models/visit.rb
```diff
@@ -19,6 +19,7 @@
 # TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 class Visit < ApplicationRecord
+  attr_accessor :set_all_insurance
```
kayla-glick, line 22: Instead of adding the `attr_accessor`, you can make the check box tag just `= check_box_tag` instead of attaching it to the form, that way it just comes across as `params[:set_all_insurance]` in the controller

app/controllers/visits_controller.rb
```diff
@@ -42,6 +42,12 @@ def update
 
     if @visit.update_attributes(visit_params)
       @visit.sub_service_request.set_to_draft unless @in_admin
+      if visit_params[:set_all_insurance] == '1'
+        insurance_qty = visit_params[:insurance_billing_qty].to_i
+        @visit_group.visits.each do |visit|
```
kayla-glick, line 47: You can use `update_all` here to update all of the records in a single query for performance

app/views/visits/_form.html.haml
jleonardw9: I'm going to leave the checkbox as is.

PR #2437, opened 2020-07-24
SJ - Epic User Validation

Stuart-Johnson:
Checking for the bad state of a project role selected for epic access, but who does not have actual epic rights, as shown through epic user validation.

[#170389425]

https://www.pivotaltracker.com/story/show/170389425


app/assets/stylesheets/protocols.scss
```diff
@@ -62,3 +62,7 @@
 /// Dashboard ///
 /////////////////
 
+#epic_rights_error_alert {
```
kayla-glick, line 65: This is cosmetic but you should be able to use bootstrap padding and margin classes instead of using a stylesheet

Stuart-Johnson: Unfortunately not really. Those only have preset sizing, none of which are correct for this situation. So unless I create a whole new entry into the variables for sizing for just this one thing, I have to use custom css (at least for the padding)

PR #2456, opened 2020-08-13
Jl - Research Quantity Reset

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/165912000


jleonardw9:
Talked to Wenjun and this whole feature is going to refactored. So this PR is no longer needed.

PR #2464, opened 2020-08-24
Jl - Protocol Update Cwf Api Bug

jleonardw9:
https://www.pivotaltracker.com/n/projects/1918597/stories/174477913

We don't need to add a protocol to the delayed job queue if it does not have any requests or fulfillment protocols.

### DIFF
--- a/app/jobs/remote_service_notifier_job.rb
+++ b/app/jobs/remote_service_notifier_job.rb
@@ -25,9 +25,16 @@ class RemoteServiceNotifierJob < Struct.new(:object_id, :object_class, :action)
 
   def self.enqueue(object_id, object_class, action)
     job = new(object_id, object_class, action)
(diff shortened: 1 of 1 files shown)


app/jobs/remote_service_notifier_job.rb
```diff
@@ -25,8 +25,18 @@ class RemoteServiceNotifierError < StandardError
 
   def self.enqueue(object_id, object_class, action)
     job = new(object_id, object_class, action)
+    should_queue = true
 
-    Delayed::Job.enqueue job, queue: 'remote_service_notifier'
+    if ((object_class == 'Study') || (object_class == 'Project'))
```
kayla-glick, line 30: Could you add a brief comment into the code for documentation in case someone stumbles on this in the future?

PR #2466, opened 2020-08-25
SAW - OnCore Study Minimum Footprint Push

SamWord:
[#171334725](https://www.pivotaltracker.com/story/show/171334725) (see story for full OnCore API specification documents)

Created an OnCore REST interface to push the minimum fields required to create a protocol in OnCore. The button to push from SPARC to OnCore is only displayed on studies in the dashboard to users who are listed in the `oncore_endpoint_access` setting if OnCore functionality is turned on. On a successful push, a flash message appears to say the study has been successfully pushed. If the protocol could not be created in SPARC, a Sweet Alert message appears saying the protocol could not be displayed and shows the response error code and message.

.env changes:

- `oncore_client_id=client_id_from_oncore` This needs to be set to the client ID that is generated by OnCore
- `oncore_client_secret=client_secret_from_oncore` This needs to be set to the client secret generated by OnCore. Since this is supposed to be a secure secret, I decided it would be better to store this as a .env variable instead of a setting in the database. Neither `oncore_client_id` or `oncore_client_secret` need to be set for institutions not using OnCore.

New setting:

- `oncore_api` should be set to the base URL of the OnCore instance for MUSC (or your institution).
- `oncore_default_department` Default OnCore field value for Primary PI's department if they don't have one.
- `oncore_default_library` Default OnCore field value for Library since nothing like this exists in SPARC.
- `oncore_default_organizational_unit` Default OnCore field value for Organizational Unit.
- `oncore_default_protocol_type` Default OnCore field value for Protocol Type.

Finally, I moved all of the spec REST request stubs that I could find to webmock.rb for organization.

Minimum footprint mapping from OnCore to SPARC:
image

### DIFF
--- a/app/lib/oncore_protocol.rb
+++ b/app/lib/oncore_protocol.rb
@@ -53,7 +53,7 @@ class OncoreProtocol
                                   protocolType: self.protocol_type
                                 })
     else
-      auth_response
+      return auth_response
     end
   end
 
@@ -66,10 +66,9 @@ class OncoreProtocol
                                 grant_type: 'client_credentials'
                               })
     if response.success?
-      token = response['access_token']
-      self.auth = "Bearer " + token
-    else
-      response
+      self.auth = "Bearer " + JSON.parse(response)['access_token']
     end
+
+    return response
   end
 end
--- a/app/lib/oncore_protocol.rb
+++ b/app/lib/oncore_protocol.rb
@@ -30,7 +30,7 @@ class OncoreProtocol
     self.title               = study.title
     self.short_title         = study.short_title
     self.library             = Setting.get_value("oncore_default_library")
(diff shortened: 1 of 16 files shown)


app/lib/oncore_protocol.rb
```diff
@@ -0,0 +1,74 @@
...
+  include HTTParty
+  base_uri Setting.get_value('oncore_api')
+
+  attr_accessor :auth, :protocol_no, :title, :short_title, :library, :department, :organizational_unit, :protocol_type
+
+  def initialize(study)
+    self.protocol_no         = "STUDY#{study.id}"
+    self.title               = study.title
+    self.short_title         = study.short_title
+    self.library             = "Non-Oncology" #default
```
kayla-glick, line 31: We should probably make the default library, organizational_unit, and protocol_type configurable. Otherwise everything else looks good to me.

PR #2512, opened 2020-09-25
SAW - Changed Epic Root for component4

SamWord:
[#174910577](https://www.pivotaltracker.com/story/show/174910577)

Changed `epic_study_root` setting to `epic_study_rsh_root` with a value of 1.2.5.2.3.4. A migration was added to update this setting. This root is no longer used in component4 segments.

Also added a setting for `epic_study_prl_root` with a value of 3.2.4.4.5 used in component4 segments of the SOAP messages sent to Epic.

### DIFF
--- a/config/defaults.json
+++ b/config/defaults.json
@@ -160,10 +160,20 @@
     "parent_value":   "true"
   },
   {
-    "key":            "epic_study_root",
-    "value":          "1.2.3.4",
-    "friendly_name":  "Epic Study Root",
+    "key":            "epic_study_rsh_root",
+    "value":          "1.2.5.2.3.4",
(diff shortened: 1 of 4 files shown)


amcates:
Looks good from a code stand point.  Once the specs are fixed I will deploy to -d

amcates:
@SamWord Will this also create the new setting in the DB or does that need to be done manually.  If manually, we need a migration.

SamWord:
@amcates I double checked with @Stuart-Johnson and he said it's fine as long as we run `rake data:import settings`.

Stuart-Johnson (approved):

PR #2727, opened 2021-07-08
Update SSR status right after SR submission while keeping email 

ying-pbrc:
…notifications in delayed job

PT:  https://www.pivotaltracker.com/n/projects/1918597/stories/178544502

Currently, when submit a service request on Step 5, the SSR status does not change until delayed job is finished; this could cause issue when delayed job is down. This change provides a solution to resolve the issue. 

Note that notifications to users and service providers are already being processed in the delayed_job:
https://github.com/sparc-request/sparc-request/blob/535895031ca9a5af30170ce0fabb7efa835046ba/app/lib/notifier_logic.rb#L136, 
https://github.com/sparc-request/sparc-request/blob/535895031ca9a5af30170ce0fabb7efa835046ba/app/lib/notifier_logic.rb#L138,
https://github.com/sparc-request/sparc-request/blob/535895031ca9a5af30170ce0fabb7efa835046ba/app/lib/notifier_logic.rb#L153.


Stuart-Johnson:
@ying-pbrc It looks like you have a spec failure related to this change.

ying-pbrc:
@Stuart-Johnson I will take a look.

Stuart-Johnson (approved):

PR #2744, opened 2021-08-09
SAW - move_service Rake Task Bug Fixes

SamWord:
[#176718864](https://www.pivotaltracker.com/story/show/176718864)

Fixed undefined variable and wrong number of [arguments errors](https://stackoverflow.com/questions/29963569/rails-upgrade-wrong-number-of-arguments-2-for-1-for-assign-attributes/29963822) preventing rake task from running and added more informative puts. Before this PR there were no puts statements indicating that any data was updated.


Stuart-Johnson (approved):
This looks good overall, I feel like Wenjun taking a close look at the results is going to be the best review XD

PR #2925, opened 2022-07-21
Add configurable demographics attributes to profile

ying-pbrc:
See details in PT: https://www.pivotaltracker.com/n/projects/1918597/stories/182491420

- New setting: displayed_demographics_fields
- Add permissible values for 'age_group', 'ethnicity', 'gender', and 'race' caetgories.
- Run  rake data:import_settings to add new setting and rake import_permissible_values to add new permissible values.


app/models/identity.rb
```diff
@@ -73,7 +73,10 @@ class Identity < ApplicationRecord
   has_many :protocol_service_requests, through: :protocols, source: :service_requests
   has_many :studies, -> { where("protocols.type = 'Study'")}, through: :project_roles, source: :protocol
 
-  cattr_accessor :current_user
```
amcates, line 76: Any reason this was removed?

ying-pbrc: @amcates that was not intentional, I will add it back then re-push to my pull request.

app/models/identity.rb
```diff
@@ -84,8 +88,9 @@ class Identity < ApplicationRecord
 
   validates :ldap_uid, uniqueness: {case_sensitive: false}, presence: true
   validates :orcid, format: { with: /\A([0-9]{4}-){3}[0-9]{3}[0-9X]\z/ }, allow_blank: true
+  # personal demographics are only required on Edit Profile page
+  validate :validate_demographics, if: Proc.new { |i| i.id == i.updater_id && i.sign_in_count > 0 }
 
-  # validates_presence_of :reason, if: :new_record?
```
amcates, line 88: Didn't notice this one earlier.  Was this also removed on purpose?

ying-pbrc: yes, it was. It was already commented out, so I deleted it. I will add it back for the record.

amcates:
@ying-pbrc Can you look in to this and resolve the conflict?  Looks like you have different permissible values.  I'm not sure which should stay in the code base.  Thanks

PR #2948, opened 2022-08-29
added new notifications for teams and specified error handling

Ito-Eta:
Recent errors in the Epic User interface due to an SSL Certificate expiration shows that there is a problem with handling errors that come from getting unexpected responses from the gateway.

In order to fix this, I specified what kind of response is expected and added the ability to send notifications to both Slack and MS Teams, depending on which an institution is using.  There is already exception notification email functionality in the application so we will continue to rely on that.

[Pivotal story](https://www.pivotaltracker.com/story/show/181922142)


app/models/epic_user.rb
```diff
@@ -35,19 +35,33 @@ def self.collection_name
 
   def self.for_identity(identity)
     begin
-      get(:viewuser, userid: identity.ldap_uid.split('@').first)
+      @result = get(:viewuser, userid: identity.ldap_uid.split('@').first)
+      @result["Silly_Name"] = @result.delete("UserID")
```
amcates: Does this need to stay in?

app/models/epic_user.rb
```diff
@@ -35,19 +35,33 @@ def self.collection_name
...
-        notifier = Slack::Notifier.new(epic_error_webhook)
-        message = I18n.t('notifier.epic_user_api_slack_error', env: Rails.env)
-        message += "\n```#{e.class}\n"
-        message += "#{e.message}\n"
-        message += "#{e.backtrace[0..5]}```"
+      slack_epic_error_webhook = Setting.get_value("epic_user_api_error_slack_webhook")
+      teams_epic_error_webhook = Setting.get_value("epic_user_api_error_teams_webhook")
+
+      message = I18n.t('notifier.epic_user_api_slack_error', env: Rails.env)
+      message += "\n```#{e.class}\n"
```
amcates, line 50: Does the same formatting work for Teams and Slack?

app/models/epic_user.rb
Ito-Eta: It absolutely does not and I thought I'd removed it.  I'll get it out.

app/models/epic_user.rb
Ito-Eta: It seems to, yes.

amcates:
@Ito-Eta  Please resolve conflicts and repush

PR #2966, opened 2022-09-12
GM - Prevent deletion of sub_service_requests record if already pushed to Fulfillment

6reg:
Users shouldn't be able to Delete a Request that has been pushed to SPARCFulfillment. The way users accomplish this is to use "Add Modify Request" and then on the dashboard screen delete the services from the cart. When the last service is deleted it triggers the SubServiceRequest to also be destroyed.
If this happens the VisitReport (and possibly other things) in SPARC Fulfillment will fail because the protocol row is storing the ID of a sub service request that will no longer exist.

[Pivotal Ticket](https://www.pivotaltracker.com/story/show/181758085)

[#181758085]

### DIFF
--- a/app/models/sub_service_request.rb
+++ b/app/models/sub_service_request.rb
@@ -72,15 +72,6 @@ class SubServiceRequest < ApplicationRecord
   after_save :update_org_tree
   after_save :update_past_status
 
-  # Users should not be able to delete a request that has been pushed
-  # to Fulfillment
-  before_destroy do
-    if self.in_work_fulfillment?
-      errors[:base] << "This request has been pushed to SPARC Fulfillment and cannot be deleted."
-      throw(:abort)
-    end
-  end
-
   scope :in_work_fulfillment, -> { where(in_work_fulfillment: true) }
   scope :imported_to_fulfillment, -> { where(imported_to_fulfillment: true) }
 
--- a/app/models/sub_service_request.rb
+++ b/app/models/sub_service_request.rb
@@ -72,6 +72,15 @@ class SubServiceRequest < ApplicationRecord
   after_save :update_org_tree
   after_save :update_past_status
 
+  # Users should not be able to delete a request that has been pushed
+  # to Fulfillment
+  before_destroy do
+    if self.in_work_fulfillment?
+      errors[:base] << "This request has been pushed to SPARC Fulfillment and cannot be deleted."
+      throw(:abort)
+    end
+  end
+
   scope :in_work_fulfillment, -> { where(in_work_fulfillment: true) }
   scope :imported_to_fulfillment, -> { where(imported_to_fulfillment: true) }
 
--- a/app/views/service_requests/cart/_cart_sub_service_requests.html.haml
+++ b/app/views/service_requests/cart/_cart_sub_service_requests.html.haml
@@ -29,13 +29,17 @@
         .list-group-item.line-item
           .row
             .col-10.d-flex.align-items-center.pr-0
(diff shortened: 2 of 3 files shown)


app/models/sub_service_request.rb
```diff
@@ -72,6 +72,15 @@ class SubServiceRequest < ApplicationRecord
   after_save :update_org_tree
   after_save :update_past_status
 
+  # Users should not be able to delete a request that has been pushed
+  # to Fulfillment
+  before_destroy do
```
jleonardw9: @6reg This is looking good so far! I pulled down your branch and did some testing and I don't think this is functioning as intended. First thing I noticed is that while the ssr is not being deleted (if `in_work_fulfillment? == true`) it's still allowing the last service to be deleted, and all the emails are still going out. Also I'm not seeing any kind of error message or anything. I think an alternate solution would be to get rid of the `before_destroy` callback, have a conditional in the view that disables the delete button if it was the last and pushed to fulfillment, and if that's the case display a tooltip on why it can't be deleted. After reading the story I noticed another fun twist to this. If any of the services under an ssr that has been pushed to fulfillment have been invoiced (there's an invoiced flag somewhere, it probably lives in fulfillment and you'll have to use the api to get to it), they also can't be deleted.

jleonardw9 (changes requested):

app/models/sub_service_request.rb
jleonardw9: Let me know if you want to pair on this!

jleonardw9: Btw, it looks like invoiced is at the fulfillment level in cwf.

jleonardw9:
@amcates Greg's only failing test is that survey thing again. Should I put a story in to clean that up and delete it? I don't think it's doing us any good if it passes locally and fails on Travis.

app/models/sub_service_request.rb
jleonardw9: @6reg Also, the way you would access that would be `Shard::Fulfillment::LineItem.where(sparc_id: line_item.id).fulfillments` or something like that, and then check is any of those fulfillments that belong to that line item are invoiced. You might have to play around with it. Look through `app/models/shard` for more examples and how they relate to our actual sparc models.

amcates:
@jleonardw9 Yes, let's remove that failing spec if it's causing issues

app/views/service_requests/cart/_cart_sub_service_requests.html.haml
```diff
@@ -29,13 +29,18 @@
...
                 = li.service.abbreviation
                 - if li.service.cpt_code.present?
                   %strong
                     = "(#{li.service.cpt_code})"
                 - unless li.service.is_available
                   = inactive_tag
             - if li.optional? && ssr.can_be_edited? && ['document_management'].exclude?(action_name)
-              .col-2.pl-0.text-right
-                = link_to icon('fas', 'trash-alt'), remove_service_service_request_path(srid: service_request.id, line_item_id: li.id), remote: true, method: :delete, class: 'btn btn-sm btn-sq btn-danger remove-service', title: t('proper.cart.remove_service'), data: { toggle: 'tooltip', disable: true }
+              - unless ssr.in_work_fulfillment && (li == ssr.line_items.last)
```
jleonardw9: @6reg I double-checked this after our discussion and I think you need `(ssr.line_items.count == 1)` instead of `(li == ssr.line_items.last)` in your check here.

jleonardw9 (changes requested):

jleonardw9 (approved):
Looks good to me!

PR #2969, opened 2022-09-19
updated question as requested

Ito-Eta:
As requested in the pivotal story, updated one of the study type questions to include new text.

pivotal story:  https://www.pivotaltracker.com/story/show/182204348


db/migrate/20220919141351_change_consent_question_in_study_questions.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeConsentQuestionInStudyQuestions < ActiveRecord::Migration[5.2]
+  def change
+    @stq = StudyTypeQuestion.find(13)
```
amcates: Finding by ID can be dangerous.   I think this might be better suited as a post deploy manual fix but maybe others can chime in

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_09_07_153451) do
+ActiveRecord::Schema.define(version: 2022_09_19_141351) do
```
amcates, line 13: Interesting that this changed.  I guess it doesn't cause any issues

amcates (changes requested):

db/migrate/20220919141351_change_consent_question_in_study_questions.rb
Ito-Eta: I just updated the migration, if that helps.  Otherwise, we can go with your idea, if necessary.  I just feel like it does need to be a migration just to make sure that anyone who pulls the code has the most current things in the database since these records are programmatically significant instead of arbitrary data.

amcates (approved):

db/migrate/20220919141351_change_consent_question_in_study_questions.rb
jleonardw9: @Ito-Eta This is great work, but I think I agree with @amcates here in that this should be a rake task instead of a migration. We should only be using migrations for schema changes (the exception here is if you need to move data around as the _result_ of a schema change). If you put a note in the release marker your rake task will get run when production is updated. Me and Sherly have a couple of them in there.

Ito-Eta: Okay.  Updated to rake task, the migration was deleted, and a note was added to the release marker.

jleonardw9 (approved):
Looks great!

PR #2971, opened 2022-09-21
JL - Line items visit subject count not correct when copying an arm

jleonardw9:
[Pivotal Story](https://www.pivotaltracker.com/n/projects/1918597/stories/165837119)

This PR fixes a bug with line items visit subject counts not being correct when copying an arm.

[#165837119]

### DIFF
--- a/app/services/arm_copier.rb
+++ b/app/services/arm_copier.rb
@@ -33,6 +33,7 @@ class ArmCopier < ApplicationService
     visit_groups.each_with_index do |group, group_index|
       if group_index == 0
         update_first_group(group)
+        update_subject_counts
       else
         create_groups_and_visits(group)
       end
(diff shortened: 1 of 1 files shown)


app/services/arm_copier.rb
```diff
@@ -64,4 +65,11 @@ def create_groups_and_visits(group)
                               effort_billing_qty: group.visits[visit_index].effort_billing_qty)
     end
   end
+
+  # Update new arm's line items visits with the correct subject count
+  def update_subject_counts
+    @copied_arm.line_items_visits.each_with_index do |line_items_visit, index|
```
amcates, line 71: Does this method have access to @copied_arm and @new_arm?

jleonardw9: @amcates I just double-checked (added come logging in the spec I wrote) and it does. I set them as instance variables in the `initialize` method.

PR #2979, opened 2022-09-28
GM - Change language and default radio

6reg:
When users add services on the SPARCRequest homepage, they are presented a pop-up screen that asks 2 different questions:

- Is this a new request AND
- Is this for a new study

The answers to these 2 questions may be different where both don't match the response button about creating a new study or going to dashboard. This request is to rephrase so only one question is being asked, revise 'No' button and customize the SPARCDashboard default filter to 'All Protocols' for 'No or Unsure' routing.

[Pivotal Ticket](https://www.pivotaltracker.com/story/show/183251824)

[#183251824]

### DIFF
--- a/app/views/service_requests/add_service.js.coffee
+++ b/app/views/service_requests/add_service.js.coffee
@@ -22,7 +22,7 @@
 ConfirmSwal.fire(
   icon: 'question'
   title: I18n.t('proper.catalog.new_request.header')
-  text: I18n.t('proper.catalog.new_request.warning')
+  text: " ",
   confirmButtonText: I18n.t('proper.catalog.new_request.yes_button')
   cancelButtonText: I18n.t('proper.catalog.new_request.no_button')
 ).then (result) ->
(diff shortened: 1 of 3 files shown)


6reg:
@jleonardw9 fyi the surveyor test that's failing on travis doesn't fail locally for me

jleonardw9 (approved):
Looks good

app/views/service_requests/add_service.js.coffee
```diff
@@ -22,7 +22,7 @@
 ConfirmSwal.fire(
   icon: 'question'
   title: I18n.t('proper.catalog.new_request.header')
-  text: I18n.t('proper.catalog.new_request.warning')
+  text: " ",
```
amcates, line 25: Should this be left empty or just removed?

6reg: It was left empty on purpose.. it's the only way I could figure how to match acceptance criteria. If the line is removed then SweetAlert auto-populates a default message at that sub-header node, but we don't want any text there.

PR #2982, opened 2022-09-28
JV - Max Length Validation Added to Protocol attribute 'short_title'

jbvanmeter:
When users previously attempted to add protocols, if the 'short_title' attribute was longer than the database-allowed 255 characters, the application would not save the new protocol record and would not indicate an error had occured. This simply adds a maximum length validation to the attribute that shows an error message if the form entry is longer than 255 characters.

https://www.pivotaltracker.com/story/show/183224023

[#183224023]


jleonardw9 (approved):
Looks good!

PR #3007, opened 2022-10-12
GM - Filter protocols default radio fix

6reg:
Currently, dashboard defaults to show "My Protocols" or "My Admin Protocols", but in cases when user is adding service to existing study then we want to default to "All Protocols" to best facilitate their search.

 [Pivotal Ticket](https://www.pivotaltracker.com/story/show/183251824)\

[#183251824]


jleonardw9 (approved):
Nice work. Good solution in `add_service.js.coffee`

PR #3009, opened 2022-10-15
GM - Multi-select documents checkbox issue fix

6reg:
There is a double checkbox showing in the select/deselect columns dropdown of the documents table. This change removes the checkbox colulmn from that menu.

[Pivotal Story](https://www.pivotaltracker.com/story/show/183429075)

[#183429075]


jleonardw9 (approved):
Great fix!

PR #3011, opened 2022-10-17
MM - update Epic StudyTypeQuestion

mam449:
Updates text for Epic StudyTypeQuestion.

Rake task included. 
`rake update_study_question_16_text`

[Pivotal](https://www.pivotaltracker.com/n/projects/1918597/stories/182736600)
[#182736600]


jleonardw9 (approved):
This looks great @mam449! One quick note, in the future any rake tasks that need to be run on production should be put in the release marker in pivotal. I can show you where it's at, and I've already added it for you.

PR #3012, opened 2022-10-19
JL - Human subjects driven logic for epic questions bug

jleonardw9:
[Pivotal Story](https://www.pivotaltracker.com/n/projects/1918597/stories/183444254)

This fixes a bug where clicking the human subjects info box on the protocols form was changing epic questions to the old text.

[#183444254]


jleonardw9:
I'm setting this back to draft. After looking at the test suite failures I think this might be intended functionality, so I think this needs some discussion.

amcates:
@jleonardw9 Is this still in draft?

jleonardw9:
@amcates This is ready! Leila said the human subjects driven logic is no longer needed, so I deleted the spec that was complaining about that.

PR #3057, opened 2022-12-21
Diplay Involved External Organizations only if use_eeeexternal_organi…

ying-pbrc:
…zations is true

https://www.pivotaltracker.com/n/projects/1918597/stories/184085253


amcates:
@ying-pbrc Can you look in to whether this is failing because of your change?

ying-pbrc:
Don't think so, it was rebuilt and all checks passed now.
________________________________

PR #3092, opened 2023-01-24
Fix undefined method 'humanize' for ActiveStorage::Filename instance error

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/184217264/comments/235210761

Export button error

### DIFF
--- a/app/controllers/funding/services_controller.rb
+++ b/app/controllers/funding/services_controller.rb
@@ -70,7 +70,7 @@ class Funding::ServicesController < ApplicationController
       documents.each do |d|
         ssr = d.share_all ? d.protocol.sub_service_requests.where(organization_id: Setting.get_value("funding_org_ids")).first : d.sub_service_requests.where(organization_id: Setting.get_value("funding_org_ids")).first
         p = ssr.protocol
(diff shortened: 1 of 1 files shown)


app/controllers/funding/services_controller.rb
```diff
@@ -70,7 +70,7 @@ def to_csv(documents)
       documents.each do |d|
         ssr = d.share_all ? d.protocol.sub_service_requests.where(organization_id: Setting.get_value("funding_org_ids")).first : d.sub_service_requests.where(organization_id: Setting.get_value("funding_org_ids")).first
         p = ssr.protocol
-        csv << [ssr.display_id, p.primary_pi.last_name_first, p.primary_pi.try(:professional_org_lookup, 'institution'), p.short_title, d.document.filename.humanize, d.updated_at.strftime('%D %l:%M %p'), PermissibleValue.get_value('status', ssr.status)]
+        csv << [ssr.display_id, p.primary_pi.last_name_first, p.primary_pi.try(:professional_org_lookup, 'institution'), p.short_title, d.document.filename, d.updated_at.strftime('%D %l:%M %p'), PermissibleValue.get_value('status', ssr.status)]
```
amcates, line 73: I think you could also do `d.document.filename.to_s.humanize` if you wanted to still use humanize but I think this fix is ok

ying-pbrc: Thanks Andrew, this is fine.

PR #3122, opened 2023-05-19
fixed billing status problem

Ito-Eta:
relevant to https://www.pivotaltracker.com/story/show/183469799


Ito-Eta:
@amcates Andrew, would you mind merging this if it's acceptable?

PR #3125, opened 2023-05-25
need to update service requester id when resubmit a service request

ying-pbrc:
https://www.pivotaltracker.com/n/projects/1918597/stories/185265894

Fix a bug where the service requester was not being updated with new submissions.

Current story will correct the sub_service_requests.service _requester_id attribute so that it's updated after the initial request.


ying-pbrc:
Hold off - per Brigette

PR #3138, opened 2023-07-04
SJ - Admin Rate History Refactor

Stuart-Johnson:
This is both a refactor for admin rate functionality, and the creation of a new "reset" functionality.

Previously, line_items could have many admin_rates, and the "correct" one was whatever the most recent one was. We wanted to move "old" admin rates somewhere else, so that a given line item would only ever have the current, correct one.

This would allow us to "reset" admin rates (be deleting them) but still keep the HISTORY of admin rates (and also, now, the record of the reset itself).

So now there is a new "admin rate changes" table, which tracks all admin rates, including making an entry when a line item is "reset."

Other changes include support for this in the modified rates report, and the addition of the reset functionality itself.


6reg (approved):

Ito-Eta (approved):
I don't see any particular problems in the code.  There's not much point in our reviewing the functionality as Brigette and Prathik will do that anyway so, approved.

PR #3144, opened 2023-07-21
this should address the problem with not being able to save the rmid …

Ito-Eta:
…change

relevant to https://www.pivotaltracker.com/story/show/185647607

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_06_06_151619) do
+ActiveRecord::Schema.define(version: 2023_06_27_001331) do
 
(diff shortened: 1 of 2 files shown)


Ito-Eta:
@amcates ready for review and merging.

db/schema.rb
```diff
@@ -33,6 +33,18 @@
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
   end
 
+  create_table "admin_rate_changes", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb3", force: :cascade do |t|
```
amcates, line 36: Curious as to why this is in your pull request

Ito-Eta: Not sure.  I didn't touch any of that.  maybe it was in code i pulled...?  but that should have been in the main branch already, yeah?  is it causing any problems?

amcates: Don't think it's really an issue but I was just curious.  I will proceed with merge

PR #3183, opened 2023-11-01
altering messages according to project rights

Ito-Eta:
https://www.pivotaltracker.com/story/show/186026699


Ito-Eta:
@amcates ready for review and merge

PR #3188, opened 2023-11-08
removed stray exclamation point

Ito-Eta:
https://www.pivotaltracker.com/story/show/186026699


Ito-Eta:
@amcates For whatever reason, Travis doesn't seem to be running but this is ready to go.  Literally, the only thing I changed is that I removed an erroneous bang on one line of code.

PR #3194, opened 2023-11-19
SJ - Service Calendar Performance 1

Stuart-Johnson:
Removing extra eager loading that was causing slowdowns. Huge improvement in load time for template tab on large studies.


app/lib/dashboard/service_calendars.rb
```diff
@@ -80,21 +80,18 @@ def self.pppv_line_items_visits_to_display(arm, service_request, sub_service_req
       if opts[:merged] && opts[:consolidated]
         if opts[:show_unchecked]
           arm.line_items_visits.
-            eager_load(:visits).
```
amcates, line 83: Pretty sure I already tested this and it breaks other parts of the code.  Travis is failing and I think it gives you a idea of other spots it's needed.  I think it might be on the Admin side that it needs `visits` to be eager loaded.

amcates (changes requested):

app/lib/dashboard/service_calendars.rb
Stuart-Johnson: So, the issue was lower down in the same query, it looks at the billing quantities on the visits. I went ahead and extended my previous billing quantity refactor to include all three billing types, and changed the query to use that instead now. All tests are passing now.

PR #3201, opened 2023-11-30
JV - Ability to Edit User Login in SPARC Admin and Toggle Off/On

jbvanmeter:
Admin users are now able to edit user logins directly from the Users table within SPARC Admin, but only if they remove "ldap_uid" from the "noneditable_identity_fields" setting array in SPARC Admin. Default behavior is to include "ldap_uid" in that array to prevent any unintentional login edits.

https://www.pivotaltracker.com/story/show/185386654

[#185386654]

### DIFF
--- a/app/controllers/admin/identities_controller.rb
+++ b/app/controllers/admin/identities_controller.rb
@@ -69,6 +69,7 @@ class Admin::IdentitiesController < Admin::ApplicationController
       :orcid,
       :credentials,
       :credentials_other,
+      :ldap_uid,
       :email,
       :era_commons_name,
       :professional_organization_id,
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -126,7 +126,7 @@ class Identity < ApplicationRecord
     return if term.blank?
 
     identity_arel = Identity.arel_table
-    attrs = [:last_name, :first_name, :email, :institution]
+    attrs = [:last_name, :first_name, :ldap_uid, :email, :institution]
 
     where (attrs
       .map { |attr| identity_arel[attr].matches("%#{term}%")}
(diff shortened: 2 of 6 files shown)


ying-pbrc:
Looks good.
Two very minor requests:
1) would you please change the Edit form header's format from "Edit User (Login ID: user_login)" to "Edit User: User Name (user@email_address)"? I put login ID in the edit header originally because it wasn't included anywhere on the form.  
2) Please update the description to include ldap_uid: #LN626 in sparc-request/config/defaults.json

Thanks!

jbvanmeter:
> Looks good. Two very minor requests:
> 
> 1. would you please change the Edit form header's format from "Edit User (Login ID: user_login)" to "Edit User: User Name (user@email_address)"? I put login ID in the edit header originally because it wasn't included anywhere on the form.
> 2. Please update the description to include ldap_uid: #LN626 in sparc-request/config/defaults.json
> 
> Thanks!

Hi Ying!

Thanks for the review. I did go ahead and include "ldap_uid" as configurable in the description (LN 626 in config/defaults.json) - so the setting description now reads, "Determines which identity fields are read-only on the edit profile page, currently only email, first_name, last_name, and ldap_uid are configurable."

Regarding the edit user form request - the header currently reads as "Edit User (Login ID: %{login})" (LN 98 in config/locales/admin.en.yml), with the %{login} interpolation insertion populating as the value in the ldap_uid database field for the given identity.

To clarify, are you wanting to change that to "Edit User: %{email}", and instead interpolate the identity's email address?

Thanks much!

ying-pbrc:
I meant to use "Edit User: %{display_name}", but thought it could also be "Edit User" for simplicity as all information are already on the form. 

Thanks!

PR #3204, opened 2023-12-01
Add additional_funding_sources setting and set it to true

6reg:
@amcates Please run `rake add_setting_for_additional_funding_sources` on -d


amcates:
Done!  I ran and it responded with `Setting created. Value set to true.`

PR #3217, opened 2023-12-12
Fix `additional_funding_sources`

6reg:
https://www.pivotaltracker.com/story/show/170740848

 - [x] Require funding source for each additional funding source record
 - [x] Funding source titles display correctly in study details view

### DIFF
--- a/app/models/additional_funding_source.rb
+++ b/app/models/additional_funding_source.rb
@@ -2,7 +2,7 @@ class AdditionalFundingSource < ApplicationRecord
   belongs_to :protocol
 
   validates :funding_source, presence: true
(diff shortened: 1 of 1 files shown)


app/models/additional_funding_source.rb
```diff
@@ -1,13 +1,15 @@
 class AdditionalFundingSource < ApplicationRecord
   belongs_to :protocol
 
+  validates :funding_source, presence: true
   validates :fundig_source_other, presence: true, if: Proc.new { |a| a.funding_source == 'Internal Funded Pilot Project' }
```
amcates: @6reg I see a typo here.  'fundig_source_other' instead of 'funding_source_other'.  Can you correct and resubmit this PR?

PR #3221, opened 2023-12-15
quick bug fix

Ito-Eta:


lib/tasks/update_protocol_with_validated_rm.rake
```diff
@@ -99,7 +99,7 @@ namespace :data do
         end
 
         # Having an eirb number implies that the protocol should have human subjects info.  Therefore, if it does not, go ahead and create a human subjects record
-        unless protocol.has_human_subject_info?
+        unless protocol.human_subjects_info.blank?
```
amcates: This should be `if` and not `unless`.   It's doing the opposite of what the other way was checking

Ito-Eta: updated

PR #3238, opened 2024-07-05
JV - Devise and Omniauth Initializer Updates for Rails 7

jbvanmeter:
Checking "Settings" values during app initialization to configure either Shibboleth or CAS authentication for OmniAuth is no longer possible in Rails 7. In order to dissuade developers from autoloading any reloadable constants during initialization (such as models like "Settings", whose attributes can be updated after initialization and initializers only run once at startup), autoloading from initializers is no longer supported.

[#185563959]


mam449 (approved):
Only notice is that the callback_url config debug is left out. Spoke with James about it.

PR #3249, opened 2025-03-24
JV - MS Teams Integration for delayed_job_monitor Rake Task

jbvanmeter:
[##188950309]


6reg (approved):

jbvanmeter:
rspec ./spec/features/surveyor/surveys/user_creates_survey_spec.rb:42 # User creates a survey surveys and sees the newly created survey

This test arbitrarily fails, failure can't be replicated when testing the application in real time and rspec failure is not consistent.

PR #3276, opened 2025-06-30
SPOSDEV-1194 Add nct# search to dashboard protocol filter

6reg:
https://sparcrequest.atlassian.net/browse/SPOSDEV-1194


mam449 (approved):
Review comments sent to Greg through Teams. Only concern from my perspective is using a "#" in a string column that may have potential to break during a query.

PR #3277, opened 2025-07-01
Sort Budget Template and Coverage Analysis reports by ssr_id

6reg:
https://sparcrequest.atlassian.net/browse/SPOSDEV-1191

### DIFF
--- a/app/models/service_request.rb
+++ b/app/models/service_request.rb
@@ -215,7 +215,11 @@ class ServiceRequest < ApplicationRecord
       end
     end
 
-    groupings
+    groupings.each do |_, group|
+      group[:line_items].sort_by!{ |li| li.sub_service_request&.ssr_id || '' }
+    end
+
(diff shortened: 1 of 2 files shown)


app/models/service_request.rb
```diff
@@ -215,7 +215,11 @@ def service_list(is_one_time_fee=nil, service_provider=nil, admin_ssr=nil)
       end
     end
 
-    groupings
+    groupings.each do |_, group|
+      group[:line_items].sort_by!{ |li| li.sub_service_request&.ssr_id || '' }
+    end
+
+    groupings.sort_by { |_, group| group[:line_items].first.sub_service_request&.ssr_id || '' }.to_h
```
mam449, line 222: Just a heads-up: if any group[:line_items] are empty, calling .first.sub_service_request (on line 222) could raise a NoMethodError as the .sub_service_request on .first would be nil. The error would raise before it falls back to ''.

